PR #17, opened 2016-08-23
Add medias with title, url and thumbnails to index

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT
| Documentation PR | sulu/sulu-docs#prnum

#### What's in this PR?

Add medias with title, url and thumbnails to index

#### Why?

So it need only to load the media when a article is changed.

### DIFF
--- a/Document/MediaCollectionOngrObject.php
+++ b/Document/MediaCollectionOngrObject.php
@@ -11,15 +11,18 @@
 
 namespace Sulu\Bundle\ArticleBundle\Document;
 
+use ONGR\ElasticsearchBundle\Annotation\Embedded;
 use ONGR\ElasticsearchBundle\Annotation\Object;
 use ONGR\ElasticsearchBundle\Annotation\Property;
+use ONGR\ElasticsearchBundle\Collection\Collection;
+use Sulu\Bundle\MediaBundle\Api\Media;
 
 /**
  * Contains the ids and display-options.
  *
  * @Object
  */
-class MediaCollectionOngrObject
+class MediaCollectionOngrObject implements \Iterator, \ArrayAccess
 {
     /**
      * @var int[]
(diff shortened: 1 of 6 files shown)


Document/MediaCollectionOngrObject.php
```diff
@@ -11,15 +11,18 @@
...
+use ONGR\ElasticsearchBundle\Collection\Collection;
+use Sulu\Bundle\MediaBundle\Api\Media;
 
 /**
  * Contains the ids and display-options.
  *
  * @Object
  */
-class MediaCollectionOngrObject
+class MediaCollectionOngrObject implements \Iterator, \ArrayAccess
```
wachterjohannes: use this (http://php.net/manual/de/class.iteratoriterator.php) as parent class which implements all the necessary functionality for you.

Document/MediaOngrObject.php
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * @var string
+     *
+     * @Property(type="string", options={"index"="not_analyzed"})
+     */
+    public $url;
+
+    /**
+     * @param Media $media
```
wachterjohannes: Use this phpdoc:

```
Set data.

@param Media $media

@return $this
```

Document/MediaOngrObject.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        $this->id = $media->getId();
+        $this->title = $media->getTitle();
+        $this->setFormats($media->getFormats());
+        $this->url = $media->getUrl();
+        $this->copyright = $media->getCopyright();
+    }
+
+    /**
+     * @return string
```
wachterjohannes: use this phpdoc:

```
Returns formats.

@return string[]
```

and update phpdoc for all the new getter and setter.

wachterjohannes:
@alexander-schranz please rebase this PR and fix the tests (:

Document/MediaCollectionOngrObject.php
alexander-schranz: needs to set it inside the construct and this will not work with ONGR.

alexander-schranz:
@wachterjohannes fixed

PR #33, opened 2016-10-04
Added tab 'all'; Fixed bug with missing selection of current tab

trickreich:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added tab 'All' in Sulu admin (can be disabled via config).

#### Why?

Fixed bug with missing selection of current tab.


Admin/ArticleJsConfig.php
```diff
@@ -33,40 +33,54 @@ class ArticleJsConfig implements JsConfigInterface
...
+            if (!array_key_exists($type, $config['types'])) {
+                $config['types'][$type] = [
                     'default' => $structure->getKey(),
                     'title' => $this->getTitle($type),
                 ];
             }
         }
 
-        return $types;
+        $config['displayTabAll'] = $this->displayTabAll;
```
wachterjohannes: move this into the initial of the var

Admin/ArticleJsConfig.php
```diff
@@ -33,40 +33,54 @@ class ArticleJsConfig implements JsConfigInterface
...
 
     /**
      * {@inheritdoc}
      */
     public function getParameters()
     {
-        $types = [];
+        $config = [];
+
+        $config['types'] = [];
```
wachterjohannes: move this into the initial of the var

DependencyInjection/Configuration.php
```diff
@@ -47,6 +47,7 @@ public function getConfigTreeBuilder()
                         ->end()
                     ->end()
                 ->end()
+                ->scalarNode('display_tab_all')->defaultTrue()->end()
```
wachterjohannes: i dont like this name (: but i cannot think of a better one. Have you any better suggestions?

wachterjohannes: also you could add her some informations with `->info('...')` (see http://symfony.com/doc/current/components/config/definition.html#documenting-the-option)

README.md
```diff
@@ -10,6 +10,19 @@ Install bundle over composer:
 composer require sulu/article-bundle
 ```
 
+Possible bundle configurations:
+
+```yml
+sulu_article:
```
wachterjohannes: if you add some info (comment above) you could use `bin/console config:dump-reference sulu_article` (where you get a commented version of the config-tree) to generate this section.

Resources/config/services.xml
```diff
@@ -16,9 +16,10 @@
             <tag name="sulu_admin.content_navigation" alias="article"/>
             <tag name="sulu.context" context="admin"/>
         </service>
-        <service id="sulu_article.js_config" class="Sulu\Bundle\ArticleBundle\Admin\ArticleJsConfig">
+        <service id="sulu_article.js_config.types" class="Sulu\Bundle\ArticleBundle\Admin\ArticleJsConfig">
```
wachterjohannes: now its not only for types. remove this suffix

Resources/public/js/components/articles/list/main.js
```diff
@@ -58,21 +59,40 @@ define(['underscore'], function(_) {
...
+                    tabItems.push(
+                        {
+                            id: type,
+                            name: types[type].title,
+                            key: type
+                        }
+                    );
+
+                    if (type === this.options.type) {
+                        tabPreselect = types[type].title;
```
wachterjohannes: cannt we use the id here? then we wont have to introduce an own var? this should be possible with `preselector: 'id', preselect: this.options.type`?

Resources/public/js/components/articles/list/main.js
```diff
@@ -168,8 +188,18 @@ define(['underscore'], function(_) {
         },
 
         typeChange: function(item) {
-            this.sandbox.emit('husky.datagrid.articles.url.update', {type: item.id});
-            this.sandbox.emit('sulu.router.navigate', 'articles:' + item.id + '/' + this.options.locale, false, false);
+            var type = null;
+            var url = 'articles';
```
wachterjohannes: i think i would introduce a underscore template like `defaults.templates.route: 'articles/<% if (!!type) { %>:<%=type%>/<% } %><%=locale%>'` which could be used here with`this.templates.route({type: this.options.type, locale: this.options.locale)`

Resources/public/js/main.js
```diff
@@ -59,14 +60,14 @@ define(['underscore', 'config'], function(_, Config) {
...
-                    }
-                });
+            app.sandbox.mvc.routes.push({
+                route: 'articles',
+                callback: function() {
+                    return app.sandbox.emit('sulu.router.navigate', 'articles/' + getLocale());
+                }
+            });
 
+            if (config.typeNames.length === 1) {
```
wachterjohannes: use yoda conditions `1 === config.typeNames.length `

Resources/public/js/main.js
```diff
@@ -81,12 +82,24 @@ define(['underscore', 'config'], function(_, Config) {
                     }
                 });
             } else {
-                app.sandbox.mvc.routes.push({
-                    route: 'articles',
-                    callback: function() {
-                        return app.sandbox.emit('sulu.router.navigate', 'articles:' + config.typeNames[0] + '/' + getLocale());
-                    }
-                });
+                if (config.displayTabAll === false) {
```
wachterjohannes: use `!config.displayTabAll`

Resources/public/js/components/articles/list/main.js
trickreich: there are three type:
* url
* position
* name

wachterjohannes: ok its not nice but ok (:

DependencyInjection/Configuration.php
trickreich: I have no better suggestion :/

Resources/public/js/components/articles/list/main.js
```diff
@@ -188,15 +193,8 @@ define(['underscore'], function(_) {
...
-            var type = null;
-            var url = 'articles';
-
-            if (!!item.key) {
-                type = item.key;
-                url += ':' + item.key;
-            }
-
-            url += '/' + this.options.locale;
+            var type = !!item.key ? item.key : null;
```
wachterjohannes: is this line necessary?

trickreich: you are right, I've removed this line

PR #56, opened 2016-12-20
Added storage to save latest type

wachterjohannes:
Main PR https://github.com/sulu/sulu/pull/3107

This PR breaks the compatability to sulu 1.4 how should we handle that? @danrot @alexander-schranz @chirimoya


alexander-schranz:
@wachterjohannes create a polyfill for this case (dont know how much work) or remove `~1.4` from the composer.json.

wachterjohannes:
i have added a small polyfill which will be registered if the missing module is not loaded.

Resources/public/js/main.js
```diff
@@ -18,6 +18,15 @@ require.config({
     }
 });
 
+if (!require.defined('services/husky/storage')) {
+
```
alexander-schranz: @wachterjohannes we should add a comment here so we now when sulu article bundle requires 1.5 we can remove this polyfill.

wachterjohannes:
@alexander-schranz fixed

danrot:
Maybe polyfill is the wrong word, because it doesn't implement the functionality at all. But that is nitpicking 😃

PR #58, opened 2016-12-22
Added methods to article-resource-item

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR allows direct access to resource-item in smart-content resource-item.

#### Why?

It is faster to use the view object that loading the document. The view-document is already there and can be extended with things that will be loaded.

### DIFF
--- a/Content/ArticleResourceItem.php
+++ b/Content/ArticleResourceItem.php
@@ -143,6 +143,26 @@ class ArticleResourceItem implements ResourceItemInterface
     }
 
     /**
+     * Returns route-path.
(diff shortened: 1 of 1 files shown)


alexander-schranz:
call it `getContent` instead of `getView` is more understandable for the frontend developer. else ACK

PR #60, opened 2016-12-22
Fixed has-next-page in article data-provider

### LINKED ISSUE #61: hasNextPage in view object is always false
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | Specific version or SHA of a commit
| Browser Version | all

#### Actual Behavior

the hasNextPage property is always false

#### Expected Behavior

to get the right value for hasNextPage from the smartcontent contenttype

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #61
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fixes the flag `hasNextPage` in `ArticleDataProvider`

### DIFF
--- a/Content/ArticleDataProvider.php
+++ b/Content/ArticleDataProvider.php
@@ -17,6 +17,7 @@ use ONGR\ElasticsearchDSL\Query\BoolQuery;
 use ONGR\ElasticsearchDSL\Query\MatchAllQuery;
 use ONGR\ElasticsearchDSL\Query\TermQuery;
 use ONGR\ElasticsearchDSL\Search;
+use ONGR\ElasticsearchDSL\Sort\FieldSort;
 use ProxyManager\Factory\LazyLoadingValueHolderFactory;
 use ProxyManager\Proxy\LazyLoadingInterface;
 use Sulu\Bundle\ArticleBundle\Document\ArticleDocument;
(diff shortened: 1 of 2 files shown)


Content/ArticleDataProvider.php
```diff
@@ -108,15 +109,22 @@ public function resolveDataItems(
...
         $uuids = [];
         /** @var ArticleViewDocumentInterface $document */
-        foreach ($this->getSearchResult($filters, $limit, $page, $pageSize) as $document) {
+        foreach ($queryResult as $document) {
             $uuids[] = $document->getUuid();
             $result[] = new ArticleDataItem($document->getUuid(), $document->getTitle(), $document);
         }
 
-        return new DataProviderResult($result, false, $uuids);
+        $count = $queryResult->count();
```
alexander-schranz: comment why count is not equal to limit

Content/ArticleDataProvider.php
```diff
@@ -134,18 +142,25 @@ public function resolveResourceItems(
...
+        foreach ($queryResult as $document) {
             $uuids[] = $document->getUuid();
             $result[] = new ArticleResourceItem(
                 $document,
                 $this->getResource($document->getUuid(), $document->getLocale())
             );
         }
 
-        return new DataProviderResult($result, false, $uuids);
+        $count = $queryResult->count();
```
alexander-schranz: looks like duplicated why not move into an own function

PR #109, opened 2017-03-21
Add caching headers and sulu twig parameters

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add caching and sulu twig parameters.

#### Why?

Articles are currently not cached.

#### BC Breaks/Deprecations

_cacheLifetime will  return the seconds instead of whole cacheLifetime array.

#### To Do

- [x] Upgrade.md
- [x] Add Tests


Routing/ArticleRouteDefaultProvider.php
```diff
@@ -85,4 +94,32 @@ public function supports($entityClass)
...
+    {
+        $cacheLifetime = $metadata->cacheLifetime;
+
+        if (!$cacheLifetime) {
+            return null;
+        }
+
+        if (!is_array($cacheLifetime)
+            || !isset($cacheLifetime['type'])
+            || !isset($cacheLifetime['value'])
```
alexander-schranz: @wachterjohannes not sure if I should check is_array, isset or just let it run into an error when it not exists.

PR #114, opened 2017-03-23
Implemented page document and API

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR introduces a `ArticlePageDocument`. This documents will be saved as childs of `ArticleDocuments`. 

#### Why?

This is the first part of the "Multi-Page" feature.

#### To Do

- [x] Unit-Tests
- [x] Indexing pages into article

### DIFF
--- a/Controller/ArticlePageController.php
+++ b/Controller/ArticlePageController.php
@@ -1,0 +1,239 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ArticleBundle\Controller;
+
+use FOS\RestBundle\Routing\ClassResourceInterface;
+use JMS\Serializer\SerializationContext;
+use Sulu\Bundle\ArticleBundle\Admin\ArticleAdmin;
+use Sulu\Bundle\ArticleBundle\Document\ArticlePageDocument;
+use Sulu\Bundle\ArticleBundle\Document\Form\ArticlePageDocumentType;
+use Sulu\Bundle\ArticleBundle\Exception\ParameterNotAllowedException;
+use Sulu\Component\Content\Form\Exception\InvalidFormException;
+use Sulu\Component\Content\Mapper\ContentMapperInterface;
+use Sulu\Component\DocumentManager\DocumentManagerInterface;
+use Sulu\Component\Rest\Exception\MissingParameterException;
+use Sulu\Component\Rest\RequestParametersTrait;
+use Sulu\Component\Rest\RestController;
+use Sulu\Component\Security\SecuredControllerInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Provides API for article-pages.
+ */
+class ArticlePageController extends RestController implements ClassResourceInterface, SecuredControllerInterface
+{
+    const DOCUMENT_TYPE = 'article_page';
+
+    use RequestParametersTrait;
+
+    /**
+     * Returns single article-page.
+     *
+     * @param string $articleUuid
+     * @param string $uuid
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function getAction($articleUuid, $uuid, Request $request)
+    {
+        $locale = $this->getRequestParameter($request, 'locale', true);
+        $document = $this->getDocumentManager()->find(
+            $uuid,
+            $locale,
+            [
+                'load_ghost_content' => true,
+                'load_shadow_content' => false,
+            ]
+        );
+
+        return $this->handleView(
+            $this->view($document)->setSerializationContext(
+                SerializationContext::create()->setSerializeNull(true)->setGroups(['defaultPage'])
+            )
+        );
+    }
+
+    /**
+     * Create article-page.
+     *
+     * @param string $articleUuid
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function postAction($articleUuid, Request $request)
+    {
+        $action = $request->get('action');
+        $document = $this->getDocumentManager()->create(self::DOCUMENT_TYPE);
+        $locale = $this->getRequestParameter($request, 'locale', true);
+        $data = $request->request->all();
+
+        $this->persistDocument($data, $document, $locale, $articleUuid);
+        $this->handleActionParameter($action, $document, $locale);
+        $this->getDocumentManager()->flush();
+
+        return $this->handleView(
+            $this->view($document)->setSerializationContext(
+                SerializationContext::create()->setSerializeNull(true)->setGroups(['defaultPage'])
+            )
+        );
+    }
+
+    /**
+     * Update article-page.
+     *
+     * @param string $articleUuid
+     * @param string $uuid
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function putAction($articleUuid, $uuid, Request $request)
+    {
+        $locale = $this->getRequestParameter($request, 'locale', true);
+        $action = $request->get('action');
+        $data = $request->request->all();
+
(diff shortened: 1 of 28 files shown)


Document/Subscriber/ArticleSubscriber.php
```diff
@@ -164,22 +173,6 @@ public function handleFlushLive(FlushEvent $event)
     }
 
     /**
-     * Indexes for article-document in live index.
-     *
-     * @param AbstractMappingEvent $event
-     */
-    public function handleIndexLive(AbstractMappingEvent $event)
```
wachterjohannes: wasnt used anymore.

Controller/ArticlePageController.php
```diff
@@ -0,0 +1,239 @@
...
+     * @throws MissingParameterException
+     * @throws ParameterNotAllowedException
+     */
+    private function persistDocument($data, $document, $locale, $articleUuid)
+    {
+        if (array_key_exists('title', $data)) {
+            throw new ParameterNotAllowedException('title', ArticlePageDocument::class);
+        }
+        if (array_key_exists('template', $data)) {
+            throw new ParameterNotAllowedException('template', ArticlePageDocument::class);
```
danrot: At least until now we always ignored extra parameters... There's the same happening for the `id` in many controllers in Sulu core... Not sure if we should start handling it here differently, resp. what would be the right way to handle this.

Document/Form/ArticleDocumentType.php
```diff
@@ -32,17 +32,12 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         // extensions
         $builder->add('extensions', TextType::class, ['property_path' => 'extensionsData']);
 
-        // TODO: Fix the admin interface to not send this junk (not required for articles)
```
danrot: Was this fix as easy as just removing these lines?

Document/Subscriber/ArticlePageSubscriber.php
```diff
@@ -0,0 +1,130 @@
...
+     * @param PersistEvent $event
+     */
+    public function setTitleOnPersist(PersistEvent $event)
+    {
+        $document = $event->getDocument();
+        if (!$document instanceof ArticlePageDocument) {
+            return;
+        }
+
+        $pageTitle = uniqid('page-', true);
```
danrot: Isn't that a strange page title? Shouldn't we fix that in a proper way? Looks like you are circumventing another problem here.

Document/Subscriber/ArticleSubscriber.php
```diff
@@ -90,7 +96,11 @@ public function handleScheduleIndex(AbstractMappingEvent $event)
     {
         $document = $event->getDocument();
         if (!$document instanceof ArticleDocument) {
-            return;
+            if (!$document instanceof ArticlePageDocument) {
+                return;
```
danrot: Just for my understanding: Is this case still possible? Or put differently: Can we still save `ArticleDocument`s, or are we always saving `ArticlePageDocuments` from now on?

Document/Subscriber/ArticleSubscriber.php
```diff
@@ -128,13 +142,11 @@ public function handleFlush(FlushEvent $event)
...
-        foreach ($this->documents as $document) {
-            $this->indexer->index(
-                $this->documentManager->find(
-                    $document['uuid'],
-                    $document['locale']
-                )
-            );
+        foreach ($this->documents as $documentData) {
+            $document = $this->documentManager->find($documentData['uuid'], $documentData['locale']);
+            $this->documentManager->refresh($document, $documentData['locale']);
```
danrot: Why is this `refresh` now necessary?

Resources/config/services.xml
```diff
@@ -122,14 +128,18 @@
             <!-- This needs to happen after the content repository has been initialized !-->
             <tag name="sulu_document_manager.initializer" priority="-127"/>
         </service>
-        <service id="sulu_article.serializer.type"
+        <service id="sulu_article.serializer.article"
```
danrot: Mind that this is a BC break. Maybe we can add the `public="false"` attribute, so that the BC break is not that bad (although it's still there, it is only less likely to occur).

Resources/config/services.xml
```diff
@@ -122,14 +128,18 @@
...
                  class="Sulu\Bundle\ArticleBundle\Document\Serializer\ArticleSubscriber">
             <argument type="service" id="sulu.content.structure_manager"/>
             <argument type="service" id="sulu.content.type_manager"/>
             <argument type="service" id="sulu_article.content.data_provider.proxy_factory"/>
 
             <tag name="jms_serializer.event_subscriber" />
         </service>
+        <service id="sulu_article.serializer.article_page"
```
danrot: Could also be private?

Resources/doc/installation.md
```diff
@@ -36,10 +36,14 @@ sulu_core:
         structure:
             default_type:
                 article: "article_default"
+                article_page: "article_default"
```
danrot: This also confuses me... Are these two types basically representing the same, and the only difference is that the `article` represents the first page only?

Controller/ArticlePageController.php
wachterjohannes: the problem here is that this has a different meaning now title is the article-title and it cannot be changed over this controller.

Document/Form/ArticleDocumentType.php
wachterjohannes: no (: the fix is:

```
     public function configureOptions(OptionsResolver $resolver)
     {
         $resolver->setDefaults(
             [
                 'allow_extra_fields' => true,
             ]
         );
     }
```

which is done since a while in this class.

Document/Subscriber/ArticleSubscriber.php
wachterjohannes: both is possible. The ArticleDocument is the first page and all the others has its own page.

Resources/config/services.xml
wachterjohannes: i will add the public=false flag.

Resources/config/services.xml
wachterjohannes: can add it also here

Resources/doc/installation.md
wachterjohannes: yes (: the configuration is a bit strange ... maybe find a better way to configure it later.

Document/Subscriber/ArticlePageSubscriber.php
wachterjohannes: it could be that the articles dont have a "pageTitle" attribute. it should not be mandatory. then we have to generate a "unique" node-name. this could be "Page 1", "Page 2" ... but that would mean that we have to update them on add and delete.

Document/Subscriber/ArticleSubscriber.php
wachterjohannes: if you add a new page during the session you have to refresh the children property to be sure that you index the newest data.

wachterjohannes:
@danrot done

Resources/config/services.xml
wachterjohannes: shit (: this does not work https://travis-ci.org/sulu/SuluArticleBundle/jobs/215473502#L491

Resources/config/services.xml
wachterjohannes: shit (: this does not work - subscriber has to be public https://travis-ci.org/sulu/SuluArticleBundle/jobs/215473502#L491

Controller/ArticlePageController.php
```diff
@@ -0,0 +1,239 @@
...
+
+        return $this->handleView($this->view(null));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSecurityContext()
+    {
+        return ArticleAdmin::SECURITY_CONTEXT;
```
alexander-schranz: 👍

PR #137, opened 2017-04-05
Fixed priorities of subscriber

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fixes the priorities to generate correct routes with pagenumber.

### DIFF
--- a/Document/ArticlePageViewObject.php
+++ b/Document/ArticlePageViewObject.php
@@ -44,6 +44,21 @@ class ArticlePageViewObject
     public $title;
 
     /**
+     * @var string
+     *
+     * @Property(
+     *     type="string",
+     *     options={
+     *         "fields":{
+     *            "raw":{"type":"string", "index":"not_analyzed"},
+     *            "value":{"type":"string"}
+     *         }
+     *     }
+     * )
+     */
+    public $routePath;
+
+    /**
      * @var int
      *
      * @Property(type="integer", options={"index"="not_analyzed"})
--- a/Document/Subscriber/RoutableSubscriber.php
+++ b/Document/Subscriber/RoutableSubscriber.php
@@ -19,8 +19,10 @@ use Sulu\Component\DocumentManager\Behavior\Mapping\ChildrenBehavior;
 use Sulu\Component\DocumentManager\Event\AbstractMappingEvent;
 use Sulu\Component\DocumentManager\Event\ConfigureOptionsEvent;
 use Sulu\Component\DocumentManager\Event\MetadataLoadEvent;
+use Sulu\Component\DocumentManager\Event\PersistEvent;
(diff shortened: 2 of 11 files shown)


Document/Subscriber/RoutableSubscriber.php
```diff
@@ -65,7 +65,7 @@ public static function getSubscribedEvents()
     {
         return [
             Events::HYDRATE => [['handleHydrate', -500]],
-            Events::PERSIST => [['handleRouteUpdate', 1], ['handleRoute', 0]],
+            Events::PERSIST => [['handleRouteUpdate', -2048], ['handleRoute', -2049]],
```
alexander-schranz: pls add some space between the priorities to allow someone to inject a validation or something

Document/Subscriber/RoutableSubscriber.php
```diff
@@ -19,15 +19,19 @@
...
 use Sulu\Component\DocumentManager\Events;
+use Sulu\Component\DocumentManager\PropertyEncoder;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
 /**
  * Handles document-manager events to create/update/remove routes.
  */
 class RoutableSubscriber implements EventSubscriberInterface
 {
+    const FIELD = 'routePath';
```
alexander-schranz: shouldn't we use a tag to mark the `routePath` and not hardcode its property name.

Document/ArticlePageViewObject.php
```diff
@@ -44,6 +44,21 @@ class ArticlePageViewObject
...
 
     /**
+     * @var string
+     *
+     * @Property(
+     *     type="string",
+     *     options={
+     *         "fields":{
+     *            "raw":{"type":"string", "index":"not_analyzed"},
+     *            "value":{"type":"string"}
```
alexander-schranz: should we use the path_hierarchy tokenizer here?

wachterjohannes: is a BC break which needs recreating the whole index. but makes sense.

wachterjohannes: `path_hierachy` is only a tokenizer and we should avoid prepending config to elasticsearch-bundle.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-pathhierarchy-tokenizer.html#_example_configuration_12

wachterjohannes:
@alexander-schranz fixed

Document/ArticlePageViewObject.php
wachterjohannes: i have created an issue for that problem https://github.com/sulu/SuluArticleBundle/issues/140

PR #172, opened 2017-05-10
Added automation to call route update asynchronously

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add a new way to update routes of articles for page-tree-routes. This is necessary if lots of articles are related to pages.

#### Why?

The save of pages (if the url was changed) will take a long time if lots of articles are related to that page.

#### Example Usage

~~~yaml
sulu_article:
    content_types:
        page_tree_route:
            update: asynchronous
~~~

#### To Do

- [x] Tests
- [x] Documentation


DependencyInjection/Configuration.php
```diff
@@ -36,15 +36,19 @@ public function getConfigTreeBuilder()
...
                         ->end()
                         ->arrayNode('page_tree_route')
                             ->addDefaultsIfNotSet()
                             ->children()
                                 ->scalarNode('template')
-                                ->defaultValue('SuluArticleBundle:Template:content-types/page-tree-route.html.twig')
+                                    ->defaultValue('SuluArticleBundle:Template:content-types/page-tree-route.html.twig')
+                                ->end()
+                                ->enumNode('update')
+                                    ->values(['synchronous', 'asynchronous', 'off'])
```
alexander-schranz: would prefer in this case us `sync` and `async` what you think @danrot @chirimoya ?

DependencyInjection/SuluArticleExtension.php
```diff
@@ -155,5 +155,13 @@ public function load(array $configs, ContainerBuilder $container)
         if (array_key_exists('SuluAutomationBundle', $bundles)) {
             $loader->load('automation.xml');
         }
+
+        if ($config['content_types']['page_tree_route']['update'] !== 'off') {
```
alexander-schranz: we should have a NullUpdater so when another bundle use the alias `sulu_article.page_tree_route.updater` and its off that it will do nothing.

DependencyInjection/SuluArticleExtension.php
```diff
@@ -155,5 +155,13 @@ public function load(array $configs, ContainerBuilder $container)
...
+            $loader->load('page_tree_update.xml');
+            $container->setAlias(
+                'sulu_article.page_tree_route.updater',
+                'sulu_article.page_tree_route.updater.' . $config['content_types']['page_tree_route']['update']
```
alexander-schranz: shouldnt we use a compiler pass to set the alias and use tagged services instead of force how the prefix of the serivce id should be?

composer.json
```diff
@@ -5,7 +5,7 @@
     "license": "MIT",
     "require": {
         "php": "^5.5 || ^7.0",
-        "sulu/sulu": "dev-develop",
+        "sulu/sulu": "dev-develop as 1.5.0",
```
alexander-schranz: this things should be removed before merging or?

composer.json
```diff
@@ -19,12 +19,16 @@
         "jackalope/jackalope-jackrabbit": "^1.2",
         "phpunit/phpunit": "^4.8 || ^5.0",
 
-        "sulu/document-manager": "@dev",
+        "sulu/document-manager": "dev-develop as 0.9.0",
         "massive/search-bundle": "@dev",
 
+        "sulu/automation-bundle": "^1.0",
```
alexander-schranz: this should be removed as it is not requirement

wachterjohannes: i need it for tests but it wont be installed if someone uses this bundle

composer.json
wachterjohannes: no before releasing

DependencyInjection/SuluArticleExtension.php
wachterjohannes: the configuration is validated strictly. when we introduce a "Pool" the validation would be quite complicated. Also, I don't think that the flexibility is necessary here.

PR #178, opened 2017-05-30
Fix caching for website articles

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Set maxAge and sharedMaxAge in the website article controller.

#### Why?

As it is require for the symfony cache to have a max and shared max age for caching.


wachterjohannes:
remove the doctrine/doctrine-bundle from composer.json then the tests should run

alexander-schranz:
@wachterjohannes done

PR #179, opened 2017-05-31
Implemented sorting pages

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT

#### What's in this PR?

This PR adds the functionallity to sort pages within an article.

#### To Do

- [x] UI
  * focus the moved row
  * flash the moved row
  * disable if only one page exists
- [x] API
- [x] Tests


alexander-schranz:
@wachterjohannes 

1. When clicking ok without change the sorting an error appears: `Missing parameter pages in Sulu\Bundle\ArticleBundle\Controller\ArticleController`

2. Draft message is not shown when change the sorting

3. Page seems to be broken because of strange urls:

The page is not published yet but the index contains strange: urls `{{ dump(pages) }}` (on website):

bildschirmfoto 2017-06-01 um 17 16 06

But the db only contains:

bildschirmfoto 2017-06-01 um 17 16 18

This how it is shown in the backend:

bildschirmfoto 2017-06-01 um 17 18 49

3. After publishing (which needs a change in the content before you can publish it) the strange urls are also in the table:

bildschirmfoto 2017-06-01 um 17 12 00
 
Code I didnt review yet.

PR #225, opened 2017-07-14
Preview serializer error

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Sometimes it seems that the parent is empty so a check is added.

#### Why?

The preview element seems to have no parent?


wachterjohannes:
closed in favor of #228

PR #235, opened 2017-08-02
Added authored and workflow-stage filter to article teaser-selection

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | requires https://github.com/sulu/sulu/pull/3464
| License | MIT

#### What's in this PR?

This PR will add the filter list to the overlay selections.


Resources/public/js/components/article-selection/main.js
```diff
@@ -103,7 +103,7 @@ define([
                         }.bind(this),
                         selectedCounter: true,
                         searchInstanceName: this.options.instanceName + '-article-search',
-                        searchFields: ['id', 'title'],
+                        searchFields: ['title', 'route_path', 'changer_full_name', 'creator_full_name', 'pages.title'],
```
alexander-schranz: author?

composer.json
```diff
@@ -5,7 +5,7 @@
     "license": "MIT",
     "require": {
         "php": "^5.5 || ^7.0",
-        "sulu/sulu": "^1.6.1",
+        "sulu/sulu": "dev-master as 1.6.3",
```
alexander-schranz: should be `^1.6.3` or `^1.6.3 || dev-master`

Resources/public/js/components/article-selection/main.js
alexander-schranz: and do we really want to be searchable for `pages.title`?

composer.json
wachterjohannes: when release 0.7 we have to change that to `^1.6.3` but now there is no 1.6.3 so this should be good.

wachterjohannes:
@alexander-schranz fixed

PR #263, opened 2017-10-11
Added exception handling when no elastic-search node is alive

### LINKED ISSUE #252: Catch ElasticSearch exceptions instead of throwing them in production
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| SuluArticleBundle Version | 0.7.0
| Sulu Version | 1.6.4

#### Actual Behavior

When there is an issue with the elasticsearch service which results in an exception being thrown, the page error 500s, taking down the whole site.

#### Expected Behavior

When there is an issue with the elasticsearch service in production environments, the exception should be logged and the search query should return null so that the rest of the site is still in operation.

#### Steps to Reproduce

Misconfigure elasticsearch parameters in production or turn the service off.

#### Possible Solutions

I had thought that a possible idea would be to dispatch an event when elastic search exceptions occur. This event could be handled by a listener that in dev/stage environments simply throws the exception, but in production has some logging and returns null. The rationale behind a listener/event implementation would be that it could easily be overwritten with custom error handling if needed. It could also be utilised in the main sulu bundle where elasticsearch maybe used for search results.

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #252 
| Related issues/PRs | none
| License | MIT

#### What's in this PR?

This PR introduces exception handling if elastic-search is not available.

#### Why?

The whole website crashes when no elastic-search is alive.

#### To Do

- [x] Add log-entry for missing elastic-search (warning)
- [x] Add documentation page

### DIFF
--- a/Content/ArticleDataProvider.php
+++ b/Content/ArticleDataProvider.php
@@ -11,6 +11,8 @@
 
 namespace Sulu\Bundle\ArticleBundle\Content;
 
+use Elasticsearch\Common\Exceptions\NoNodesAvailableException;
+use ONGR\ElasticsearchBundle\Result\DocumentIterator;
 use ONGR\ElasticsearchBundle\Service\Manager;
 use ONGR\ElasticsearchDSL\Query\Compound\BoolQuery;
 use ONGR\ElasticsearchDSL\Query\MatchAllQuery;
@@ -19,6 +21,7 @@ use ONGR\ElasticsearchDSL\Search;
 use ONGR\ElasticsearchDSL\Sort\FieldSort;
 use ProxyManager\Factory\LazyLoadingValueHolderFactory;
 use ProxyManager\Proxy\LazyLoadingInterface;
+use Psr\Log\LoggerInterface;
 use Sulu\Bundle\ArticleBundle\Document\ArticleDocument;
 use Sulu\Bundle\ArticleBundle\Document\ArticleViewDocumentInterface;
 use Sulu\Bundle\WebsiteBundle\ReferenceStore\ReferenceStoreInterface;
@@ -71,6 +74,11 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
     protected $defaultLimit;
 
     /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * @param Manager $searchManager
      * @param DocumentManagerInterface $documentManager
      * @param LazyLoadingValueHolderFactory $proxyFactory
@@ -78,6 +86,7 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
      * @param ArticleResourceItemFactory $articleResourceItemFactory
      * @param string $articleDocumentClass
      * @param int $defaultLimit
+     * @param LoggerInterface $logger
      */
     public function __construct(
         Manager $searchManager,
@@ -86,7 +95,8 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
         ReferenceStoreInterface $referenceStore,
         ArticleResourceItemFactory $articleResourceItemFactory,
         $articleDocumentClass,
-        $defaultLimit
+        $defaultLimit,
+        LoggerInterface $logger = null
     ) {
         $this->searchManager = $searchManager;
         $this->documentManager = $documentManager;
@@ -95,6 +105,7 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
         $this->articleResourceItemFactory = $articleResourceItemFactory;
         $this->articleDocumentClass = $articleDocumentClass;
         $this->defaultLimit = $defaultLimit;
+        $this->logger = $logger;
     }
 
     /**
@@ -162,7 +173,7 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
             $result[] = new ArticleDataItem($document->getUuid(), $document->getTitle(), $document);
         }
 
-        return new DataProviderResult($result, $this->hasNextPage($queryResult, $limit, $page, $pageSize), $uuids);
+        return new DataProviderResult($result, $this->hasNextPage($queryResult, $limit, $page, $pageSize));
     }
 
     /**
@@ -191,7 +202,7 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
             $result[] = $this->articleResourceItemFactory->createResourceItem($document);
         }
 
-        return new DataProviderResult($result, $this->hasNextPage($queryResult, $limit, $page, $pageSize), $uuids);
+        return new DataProviderResult($result, $this->hasNextPage($queryResult, $limit, $page, $pageSize));
     }
 
     /**
@@ -239,7 +250,7 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
         $repository = $this->searchManager->getRepository($this->articleDocumentClass);
         $search = $this->createSearch($repository->createSearch(), $filters, $locale);
         if (!$search) {
-            return new \ArrayIterator([]);
+            return new DocumentIterator([], $this->searchManager);
         }
 
         $this->addPagination($search, $pageSize, $page, $limit);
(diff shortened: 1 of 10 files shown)


Content/ArticleDataProvider.php
```diff
@@ -86,7 +95,8 @@ public function __construct(
         ReferenceStoreInterface $referenceStore,
         ArticleResourceItemFactory $articleResourceItemFactory,
         $articleDocumentClass,
-        $defaultLimit
+        $defaultLimit,
+        LoggerInterface $logger = null
```
alexander-schranz: I personally would go the PSR way of injecting the logger over the `setLogger` using the LoggerAwareTrait:

https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md

Content/ArticleDataProvider.php
```diff
@@ -95,6 +105,7 @@ public function __construct(
         $this->articleResourceItemFactory = $articleResourceItemFactory;
         $this->articleDocumentClass = $articleDocumentClass;
         $this->defaultLimit = $defaultLimit;
+        $this->logger = $logger;
```
alexander-schranz: when use the PSR way set the logger to `new NullLogger()` in the construct so it throws no error when the logger isn't injected.

Document/Repository/ArticleViewDocumentRepository.php
```diff
@@ -57,19 +59,28 @@ class ArticleViewDocumentRepository
...
      * @param string $articleDocumentClass
      * @param array $searchFields
+     * @param LoggerInterface|null $logger
      */
     public function __construct(
         Manager $searchManager,
         $articleDocumentClass,
-        array $searchFields
+        array $searchFields,
+        LoggerInterface $logger = null
```
alexander-schranz: same: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md

Markup/ArticleLinkProvider.php
```diff
@@ -48,21 +50,29 @@ class ArticleLinkProvider implements LinkProviderInterface
...
      * @param string $articleViewClass
+     * @param LoggerInterface|null $logger
      */
     public function __construct(
         Manager $liveManager,
         Manager $defaultManager,
         array $types,
-        $articleViewClass
+        $articleViewClass,
+        LoggerInterface $logger = null
```
alexander-schranz: same https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md

Sitemap/ArticleSitemapProvider.php
```diff
@@ -37,13 +40,23 @@ class ArticleSitemapProvider implements SitemapProviderInterface
...
-    public function __construct(Manager $manager, DocumentFactoryInterface $documentFactory)
-    {
+    public function __construct(
+        Manager $manager,
+        DocumentFactoryInterface $documentFactory,
+        LoggerInterface $logger = null
+    ) {
         $this->manager = $manager;
         $this->documentFactory = $documentFactory;
+        $this->logger = $logger;
```
alexander-schranz: same https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md

Teaser/ArticleTeaserProvider.php
```diff
@@ -43,15 +45,26 @@ class ArticleTeaserProvider implements TeaserProviderInterface
...
      * @param $articleDocumentClass
+     * @param LoggerInterface|null $logger
      */
-    public function __construct(Manager $searchManager, TranslatorInterface $translator, $articleDocumentClass)
-    {
+    public function __construct(
+        Manager $searchManager,
+        TranslatorInterface $translator,
+        $articleDocumentClass,
+        LoggerInterface $logger = null
```
alexander-schranz: same https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md

Sitemap/ArticleSitemapProvider.php
```diff
@@ -117,7 +130,15 @@ private function getBulk(Repository $repository, $from, $size)
             ->setFrom($from)
             ->setSize($size);
 
-        return $repository->findDocuments($search);
+        try {
+            return $repository->findDocuments($search);
+        } catch (NoNodesAvailableException $exception) {
+            if ($this->logger) {
```
alexander-schranz: not needed when setting the logger to `new NullLogger()` in the construct.

Content/ArticleDataProvider.php
```diff
@@ -249,7 +260,15 @@ private function getSearchResult(array $filters, $limit, $page, $pageSize, $loca
             $this->appendSortBy($filters['sortBy'], $sortMethod, $search);
         }
 
-        return $repository->findDocuments($search);
+        try {
+            return $repository->findDocuments($search);
+        } catch (NoNodesAvailableException $exception) {
+            if ($this->logger) {
```
alexander-schranz: not needed when setting the logger to `new NullLogger()` in the construct

Content/ArticleSelectionContentType.php
```diff
@@ -83,9 +93,19 @@ public function getContentData(PropertyInterface $property)
         $search = $repository->createSearch();
         $search->addQuery(new IdsQuery($this->getViewDocumentIds($value, $locale)));
 
+        try {
+            $documents = $repository->findDocuments($search);
+        } catch (NoNodesAvailableException $exception) {
+            if ($this->logger) {
```
alexander-schranz: not needed when logger is `NullLogger` when no logger is given.

Document/Repository/ArticleViewDocumentRepository.php
```diff
@@ -92,7 +103,15 @@ public function findRecent($excludeUuid = null, $limit = self::DEFAULT_LIMIT, ar
 
         $search->addSort(new FieldSort('authored', FieldSort::DESC));
 
-        return $this->repository->findDocuments($search);
+        try {
+            return $this->repository->findDocuments($search);
+        } catch (NoNodesAvailableException $exception) {
+            if ($this->logger) {
```
alexander-schranz: same use `NullLogger`

Document/Repository/ArticleViewDocumentRepository.php
```diff
@@ -121,7 +140,15 @@ public function findSimilar($uuid, $limit = self::DEFAULT_LIMIT, array $types =
             )
         );
 
-        return $this->repository->findDocuments($search);
+        try {
+            return $this->repository->findDocuments($search);
+        } catch (NoNodesAvailableException $exception) {
+            if ($this->logger) {
```
alexander-schranz: use `NullLogger`

Markup/ArticleLinkProvider.php
```diff
@@ -101,7 +111,15 @@ public function preload(array $hrefs, $locale, $published = true)
             $repository = $this->defaultManager->getRepository($this->articleViewClass);
         }
 
-        $documents = $repository->findDocuments($search);
+        try {
+            $documents = $repository->findDocuments($search);
+        } catch (NoNodesAvailableException $exception) {
+            if ($this->logger) {
```
alexander-schranz: use `NullLogger`

Teaser/ArticleTeaserProvider.php
```diff
@@ -169,19 +179,29 @@ public function find(array $ids, $locale)
         $search = $repository->createSearch();
         $search->addQuery(new IdsQuery($articleIds));
 
+        try {
+            $documents = $repository->findDocuments($search);
+        } catch (NoNodesAvailableException $exception) {
+            if ($this->logger) {
```
alexander-schranz: use `NullLogger`

Content/ArticleDataProvider.php
```diff
@@ -249,7 +257,13 @@ private function getSearchResult(array $filters, $limit, $page, $pageSize, $loca
             $this->appendSortBy($filters['sortBy'], $sortMethod, $search);
         }
 
-        return $repository->findDocuments($search);
+        try {
+            return $repository->findDocuments($search);
+        } catch (NoNodesAvailableException $exception) {
+            $this->logger->error($exception->getMessage());
```
alexander-schranz: use `critical` log

alexander-schranz:
maybe we should add a `view` parameter in the content types that he can show a error message in the frontend when the exception is catched.

alexander-schranz:
ping @wachterjohannes

wachterjohannes:
ping @sulu/core-team

wachterjohannes:
we have decided to not handle this exceptions. the reason is that there is no "correct" way for this. we see elasticsearch as a hard dependency for the application like mysql which also will bring the application to crash if it is not there.

PR #291, opened 2017-12-07
Fix drop index on none exist and remove ArticleIndexBuilder dependency on cache builder

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

The cache clear need to be removed for 1.6.9

#### Why?

This was removed for symfony 3.4 in sulu/sulu.

### DIFF
--- a/Builder/ArticleIndexBuilder.php
+++ b/Builder/ArticleIndexBuilder.php
@@ -32,7 +32,7 @@ class ArticleIndexBuilder extends SuluBuilder
      */
     public function getDependencies()
     {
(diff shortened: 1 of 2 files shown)


trickreich:
@alexander-schranz We still need this?

alexander-schranz:
@trickreich its just not longer critical since the fix in sulu 1.6.10. But the changes can be done.

PR #293, opened 2017-12-26
Improve "My articles" translations NL

mark-gerarts:
The old translation wasn't correct in Dutch, it translated roughly as "Me article". This updates it to "My articles"


wachterjohannes:
@mark-gerarts thanks for improvement :)

PR #316, opened 2018-03-20
Added smart content filter by structure type (article.xml -> key attribute)

dmitry-pro:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added structureType filtering ability for articles smart-content provider.
It allows to filter articles not only by type, but also by article template `key` tag value.

`
<?xml version="1.0" ?>
<template xmlns="http://schemas.sulu.io/template/template"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xmlns:xi="http://www.w3.org/2001/XInclude"
          xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/template-1.0.xsd">

    <key>special_company_blog_article</key>
    ...
`

#### Why?

It is very convenient if we want to display list of same article types on different webspaces.

#### Example Usage

Simple use `structureTypes` param to filter articles by it:

`
                <properties>
                    <property name="articles" type="smart_content" mandatory="true">
                        <meta>
                            <title lang="en">Articles</title>
                            <title lang="uk">Статті</title>
                        </meta>

                        <params>
                            <param name="provider" value="articles"/>
                            <param name="types" value="blog"/>
                            <param name="structureTypes" value="special_company_blog_article"/>
                            <param name="max_per_page" value="3"/>
                        </params>
                    </property>
                </properties>
`
#### To Do

- [ ] Add documentation page

### DIFF
--- a/Content/ArticleDataProvider.php
+++ b/Content/ArticleDataProvider.php
@@ -150,6 +150,7 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
         $pageSize = null
     ) {
         $filters['types'] = $this->getTypesProperty($propertyParameter);
+        $filters['structureTypes'] = $this->getStructureTypesProperty($propertyParameter);
         $filters['excluded'] = $this->getExcludedFilter($filters, $propertyParameter);
 
         $queryResult = $this->getSearchResult($filters, $limit, $page, $pageSize, $options['locale']);
@@ -175,6 +176,7 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
         $pageSize = null
     ) {
         $filters['types'] = $this->getTypesProperty($propertyParameter);
+        $filters['structureTypes'] = $this->getStructureTypesProperty($propertyParameter);
         $filters['excluded'] = $this->getExcludedFilter($filters, $propertyParameter);
 
         $queryResult = $this->getSearchResult($filters, $limit, $page, $pageSize, $options['locale']);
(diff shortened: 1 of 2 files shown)


Content/ArticleDataProvider.php
```diff
@@ -150,6 +150,7 @@ public function resolveDataItems(
         $pageSize = null
     ) {
         $filters['types'] = $this->getTypesProperty($propertyParameter);
+        $filters['structureTypes'] = $this->getStructureTypesProperty($propertyParameter);
```
alexander-schranz: I would call it templates /cc @wachterjohannes what do you think. Can you have a look at this PR?

wachterjohannes: @alexander-schranz I am not sure - but I think `structureType` is "more" correct - because we use this nearly on all places in the PHP code.

alexander-schranz: but we use always the term `template` for configuration and in the apis so I would go to `templates` here also and for this who create the xml's it will always be a template and not a structure. So for me structureType is really false here for me.

dmitry-pro: @wachterjohannes Updated doc page, is the text ok?

dmitry-pro: For my opinion `template` looks more "organic", but we already use `structureType` name for it in Smart Content structure, so it will be breaking if we change it everywhere, and will look messy if we change it only in single place.
Probably, it would be good to rename this key to `template` everywhere in next release, but it's up to You to decide.

wachterjohannes:
@dmitry-pro thanks for the contribution

dmitry-pro:
@wachterjohannes no sweat 👍

PR #330, opened 2018-06-08
Fixed article page publish by hydrating parent-document

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT

#### What's in this PR?

ArticlePages cannot be published because ArticleDocument (parent of article-page) is a proxy and therfor the route cannot be found.


Document/Subscriber/ArticlePageSubscriber.php
```diff
@@ -88,6 +89,9 @@ public function __construct(
     public static function getSubscribedEvents()
     {
         return [
+            Events::HYDRATE => [
+                ['setParentOnHydrate', 1],
```
trickreich: shouldn't it be something like `loadParentOnHydrate` or `resolveParentOnHydrate`?

wachterjohannes: the naming in this file is always `setSomthingOnEvent`

PR #353, opened 2018-10-19
Replace ONGR Collection in favour of Doctrine ArrayCollection

AcelisWeaven:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This issue fixes an error occurring with a recent version of ONGR.
After installing the bundle and going to the articles list, the following error occurs:
```
Type error: Argument 1 passed to Sulu\Bundle\ArticleBundle\Document\ArticleViewDocument::setPages() must be an instance of ONGR\ElasticsearchBundle\Collection\Collection, instance of ONGR\ElasticsearchBundle\Result\ObjectIterator given, called in /application/vendor/ongr/elasticsearch-bundle/Result/Converter.php on line 134
```

This happen when Sulu access the URL `/admin/api/articles?fields=id%2CtypeTranslation%2Ctitle%2CchangerFullName%2CauthorFullName%2Cchanged%2Cauthored&limit=20&locale=en&sortBy=typeTranslation&sortOrder=asc&type=article`

ongr/elasticsearch-bundle version: v5.0.7
ES version: 5.0.2

#### Why?

A workaround has been found [in this Sylius issue](https://github.com/Sylius/SyliusElasticSearchPlugin/issues/90), where the `ONGR\ElasticsearchBundle\Collection\Collection` references were replaced by `Doctrine\Common\Collections\Collection`.

This replacement seems to fix the issue, but I don't know if it could have any side-effect.

### DIFF
--- a/Document/ArticleViewDocument.php
+++ b/Document/ArticleViewDocument.php
@@ -15,7 +15,7 @@ use ONGR\ElasticsearchBundle\Annotation\Document;
 use ONGR\ElasticsearchBundle\Annotation\Embedded;
 use ONGR\ElasticsearchBundle\Annotation\Id;
 use ONGR\ElasticsearchBundle\Annotation\Property;
-use Doctrine\Common\Collections\Collection;
+use ONGR\ElasticsearchBundle\Collection\Collection;
 
 /**
  * Indexable document for articles.
--- a/Document/ArticleViewDocument.php
+++ b/Document/ArticleViewDocument.php
@@ -15,7 +15,7 @@ use ONGR\ElasticsearchBundle\Annotation\Document;
 use ONGR\ElasticsearchBundle\Annotation\Embedded;
 use ONGR\ElasticsearchBundle\Annotation\Id;
 use ONGR\ElasticsearchBundle\Annotation\Property;
-use ONGR\ElasticsearchBundle\Collection\Collection;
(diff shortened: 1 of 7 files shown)


trickreich:
@wachterjohannes You've seen this?

wachterjohannes:
@trickreich thanks for the hint :)

@AcelisWeaven we also have thought about this change - but the problem is, that this would break the compatibility to older versions of ONGR-Bundle. And i think this is also the reason of the failing test cases.

AcelisWeaven:
@wachterjohannes Thanks for the hint, turns out the error on my side. I was still on 0.7.1 while making those changes, and I didn't test my PR on 1.0.0-RC7.
In my project, I upgraded my requirement for SuluArticleBundle, and ongr/elasticsearch-bundle was automatically downgraded from v5.0.7 to v5.0.5, which solved my issue.

If anyone have the same issue: `composer require sulu/article-bundle 1.0.0-RC7 ongr/elasticsearch-bundle ~5.0`.

I'm going to close this PR since this actually fixes nothing.
In the meantime, maybe a 0.7.2 version with an ongr ^5.0.6 conflict entry (like in 1.0.0-RC) would to the trick? What do you think?

PR #355, opened 2018-11-12
Fixed incompatibility with ongr/elasticsearch 5.0.6

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added support ongr newest version after a bc break in ongr ( https://github.com/ongr-io/ElasticsearchBundle/pull/777 )

#### Why?

To support needed fixes of 5.0.6.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,8 +24,7 @@
         "php-task/php-task": "^1.2"
     },
     "conflict": {
-        "php": "^7.2",
-        "ongr/elasticsearch-bundle": "^5.2.0"
+        "php": "^7.2"
     },
     "suggest": {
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -25,7 +25,7 @@
     },
     "conflict": {
         "php": "^7.2",
-        "ongr/elasticsearch-bundle": "^5.0.6"
+        "ongr/elasticsearch-bundle": "^5.2.0"
```
trickreich: Check if this is still needed

Issue #304, opened 2018-02-05
Elasticsearch ^6.0 compatibility


wachterjohannes:
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| SuluArticleBundle Version | all
| Sulu Version | ^1.6

#### Actual Behavior

ongr-io/ElasticsearchBundle ~~does not support PHP `^7.2` (https://github.com/ongr-io/ElasticsearchBundle/issues/832)~~ and elasticsearch `^6.0` (https://github.com/ongr-io/ElasticsearchBundle/issues/828)

#### Expected Behavior

This should be supported.

#### Possible Solutions

* Remove dependency to ongr-io/ElasticsearchBundle (rewrite of indexing - e.g. using raw elasticsearch library)
* Find a better solution for the performance problems and remove elasticsearch dependency

mareksuscak:
What's the current status? I'm a little confused because according to [this comment](https://github.com/ongr-io/ElasticsearchBundle/issues/828#issuecomment-451946214) the issue got resolved in ElasticsearchBundle 5.2.0 but despite having installed version 5.2.4 and SuluArticleBundle 1.0.1, I'm still seeing weird autoloading errors.

wachterjohannes:
@mareksuscak this issue is only resolved partly - the bundle is now ready for 7.2 but not ES 6.0

mareksuscak:
Which version is ready for PHP 7.2 though? The latest one (i.e. 1.0.1)? Or the development version that's not been released yet? 1.0.1 didn't work w/ PHP 7.2 and Elasticsearch server 5.x.

wachterjohannes:
all versions ^1.0 should be ready for php 7.2 and ES ^5.0 - what is you exact problem? do you have any error message?

mareksuscak:
There was a problem w/ the autoloader. Specifically, it was looking for `Object` class in [this file](https://github.com/ongr-io/ElasticsearchBundle/blob/master/Annotation/Object.php) and said the file was there but the class was not found.

YetiCGN:
@mareksuscak I think you need to use the new annotation `@ObjectType` instead of `@Object` in your own code.

mareksuscak:
We don't use that annotation in our own code. Just tried it again w/ the same result. However, upon manually removing the contents of the cache folder, it started working. I'm pretty sure I ran `cache:clear` command before.

alexander-schranz:
[...]

PR #357, opened 2018-11-15
Compatibility with PHP 7.2 by updating elasticsearch-bundle dependency

YetiCGN:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | #304 (the "PHP 7.2" part)
| License | MIT

#### What's in this PR?

Ensures compatiblity with PHP 7.2 by raising ongr/elasticsearch-bundle version to ^5.2, which is certified to work with PHP 7.2. Also contains the necessary renaming of the "Object" annotation to "ObjectType" (object is a reserved keyword starting in PHP 7.2)

#### Why?

SuluArticleBundle should be able to work with the most current supported PHP version.

#### Example Usage

Just do it.

#### BC Breaks/Deprecations

None that I am aware of

#### To Do

?

### DIFF
--- a/Document/ArticlePageViewObject.php
+++ b/Document/ArticlePageViewObject.php
@@ -11,13 +11,13 @@
 
 namespace Sulu\Bundle\ArticleBundle\Document;
 
-use ONGR\ElasticsearchBundle\Annotation\Object;
+use ONGR\ElasticsearchBundle\Annotation\ObjectType;
 use ONGR\ElasticsearchBundle\Annotation\Property;
 
 /**
  * Contains page information.
  *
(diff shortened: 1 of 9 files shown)


trickreich:
@YetiCGN Thanks for the contribution! :+1: 
Currently we are support elasticsearch 2 also.
This is the reason why we haven't added such a bc break to our bundle.
We need to make a plan how we can support PHP 7.2 as fast as possible.
I'll connect back to tyou.

PR #365, opened 2018-12-12
Fix php 7.3 compatibility

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Test sulu article bundle on php 7.3 and master sulu/sulu.

#### Why?

Php 7.3 was released last week and should be checked.


alexander-schranz:
Fixed in https://github.com/sulu/SuluArticleBundle/pull/368/files

PR #371, opened 2019-01-31
Fix getUuid for article-bridge used in content resolver

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT

#### What's in this PR?

This PR add the `getUuid` function to `ArticleBridge` because this function is used for resolving content.

### DIFF
--- a/Document/Structure/ArticleBridge.php
+++ b/Document/Structure/ArticleBridge.php
@@ -24,11 +24,32 @@ class ArticleBridge extends StructureBridge
     private $webspaceKey = null;
 
     /**
+     * @var string
+     */
+    private $uuid;
+
+    /**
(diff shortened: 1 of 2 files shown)


Document/Structure/ArticleBridge.php
```diff
@@ -31,6 +36,20 @@ public function getView()
         return $this->structure->view;
     }
 
+    public function getUuid()
+    {
+        if ($this->document) {
```
alexander-schranz: Can you add a comment when this happens and why?

PR #381, opened 2019-04-17
Add the option to give permissions per article type

### LINKED ISSUE #373: Set user permissions to specific article type(s)
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| SuluArticleBundle Version | Specific version or SHA of a commit
| Sulu Version | Specific version or SHA of a commit

#### Actual Behavior
You only could give permission to have access to articles or not.

#### Expected Behavior
Assign specific article type(s) to an user (via roles)

#### Steps to Reproduce

-

#### Possible Solutions

-

TisLars:
With this implementation it is possible to give permissions per article type. When the specific user has no permission to view the type, the item is not available in the overview.

| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #373
| License | MIT

#### What's in this PR?

The ArticleAdmin loops through all article types and for each type, adds a new permission: Article type access. When permission is not granted, these article types will not show up in the navigation bar.

#### Why?

Gives admins the ability to restrict users to certain article types. 

An example could be a user that can log into sulu but only has access to edit blog articles. All other sulu functionalities would be hidden from this user group, and only blog articles can be shown.

#### Example Usage

config.yml
~~~yml
 # Display tab 'all' in list view
display_tab_all: false
~~~
Adding this to the sulu_article config removes the all tab, so the users in this role have no access to different article types. 

#### BC Breaks/Deprecations

The only non-bc is the default of the articles is set to false, so existing CMS users need to set their permissions to true to see the article types.

#### To Do (optionals)

- [ ] Add documentation page
- [ ] Create getters and setters for the security context in ArticleAdmin
- [ ] Add an extra check for article permission but no type permission for BC


Admin/ArticleAdmin.php
```diff
@@ -11,32 +11,65 @@
...
+        array $securityContext
+    )
     {
         $rootNavigationItem = new NavigationItem($title);
         $section = new NavigationItem('navigation.modules');
         $section->setPosition(20);
+        $types = [];
+        $this->securityContext = $securityContext;
+
+        foreach ($structureManager->getStructures('article') as $key => $structure) {
```
wachterjohannes: you should do this inside the `getSecurityContexts` - else this "heavy" operation will be done for all requests

Admin/ArticleAdmin.php
```diff
@@ -11,32 +11,65 @@
...
+            $type = $this->getType($structure->getStructure());
+            if (!array_key_exists($type, $types)) {
+                $types[$type] = [
+                    'type' => $structure->getKey(),
+                ];
+            }
+            $this->securityContext[self::SECURITY_CONTEXT . ' ' . $type] = [
```
wachterjohannes: i think we should use an `_` for the concatination

Admin/ArticleAdmin.php
```diff
@@ -50,6 +83,7 @@ public function __construct(SecurityCheckerInterface $securityChecker, $title)
         }
 
         $this->setNavigation(new Navigation($rootNavigationItem));
+
```
wachterjohannes: remove empty line

Admin/ArticleAdmin.php
```diff
@@ -11,32 +11,65 @@
...
+     * @param array $securityContext
      */
-    public function __construct(SecurityCheckerInterface $securityChecker, $title)
+    public function __construct
+    (
+        SecurityCheckerInterface $securityChecker,
+        StructureManagerInterface $structureManager,
+        $title,
+        array $securityContext
+    )
```
wachterjohannes: the `{` should be in the same line as `)`

Admin/ArticleAdmin.php
```diff
@@ -11,32 +11,65 @@
...
+                    'type' => $structure->getKey(),
+                ];
+            }
+            $this->securityContext[self::SECURITY_CONTEXT . ' ' . $type] = [
+                        PermissionTypes::VIEW,
+                        PermissionTypes::ADD,
+                        PermissionTypes::EDIT,
+                        PermissionTypes::DELETE,
+                        PermissionTypes::LIVE,
+                    ];
```
wachterjohannes: the indentation is wrong:

```
$this->securityContext[self::SECURITY_CONTEXT . ' ' . $type] = [
    ...
];
```

Admin/ArticleAdmin.php
```diff
@@ -65,18 +99,21 @@ public function getJsBundleName()
...
+                            PermissionTypes::ADD,
+                            PermissionTypes::EDIT,
+                            PermissionTypes::DELETE,
+                            PermissionTypes::LIVE,
+                        ],
                     ],
-                ],
-            ],
-        ];
+                    'Article type access' => $this->securityContext,
```
wachterjohannes: would use `Article Types` as name here

Admin/ArticleAdmin.php
```diff
@@ -65,18 +99,21 @@ public function getJsBundleName()
...
-        return [
-            'Sulu' => [
-                'Global' => [
-                    self::SECURITY_CONTEXT => [
-                        PermissionTypes::VIEW,
-                        PermissionTypes::ADD,
-                        PermissionTypes::EDIT,
-                        PermissionTypes::DELETE,
-                        PermissionTypes::LIVE,
+        $contexts = [
```
wachterjohannes: return this array without an additional variable

Admin/ArticleAdmin.php
```diff
@@ -11,32 +11,65 @@
...
      * @param string $title
+     * @param array $securityContext
      */
-    public function __construct(SecurityCheckerInterface $securityChecker, $title)
+    public function __construct
+    (
+        SecurityCheckerInterface $securityChecker,
+        StructureManagerInterface $structureManager,
+        $title,
+        array $securityContext
```
wachterjohannes: do you need this variable?

Resources/public/js/components/articles/list/main.js
```diff
@@ -127,6 +128,11 @@ define([
 
                 // add tab item for each type
                 _.each(typeNames, function(type) {
+                    var permissions = Config.get('sulu_security.contexts')['sulu.modules.articles ' + type];
+                    if (permissions['view'] === false) {
```
wachterjohannes: 👍 for the JS changes we need a `grunt build` in the root of the bundle to build the dist (minified) files

wachterjohannes: and i think we should move this check to https://github.com/sulu/SuluArticleBundle/blob/develop/Resources/public/js/main.js#L53 to make sure that article types without permision never used inside the frontend

TisLars: > 👍 for the JS changes we need a `grunt build` in the root of the bundle to build the dist (minified) files

I do not completely understand this comment. Do you want to see a change?

wachterjohannes: the bundle contains the build-files for JS admin in prod mode - to rebuild them run `npm install` and `grund build` in the root of the bundle

wachterjohannes:
@guuse everything looks good /cc @alexander-schranz 

whats now missing is that we have to add a `UPGRADE.md` section to say that this new security-contexts has to be set in the admin UI

PR #429, opened 2020-01-17
Add getters for the author's id and full name

DavidFeller:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds getter methods to Sulu\Bundle\ArticleBundle\Content\ArticleResourceItem to allow for displaying the author's id and full name in article lists.

#### Why?

At the moment there is no way to display the mentioned information a article list.

### DIFF
--- a/Content/ArticleResourceItem.php
+++ b/Content/ArticleResourceItem.php
@@ -222,7 +222,6 @@ class ArticleResourceItem implements ResourceItemInterface
         return $this->getUuid();
     }
 
-
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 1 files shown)


Content/ArticleResourceItem.php
```diff
@@ -221,4 +221,21 @@ public function getId()
     {
         return $this->getUuid();
     }
+
```
wachterjohannes: @DavidFeller i have no idea why pretty is not fixing this on your local machine - i would remove this line 😂

PR #436, opened 2020-02-18
Add tab-view for article-types

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | https://github.com/sulu/sulu/pull/5066
| License | MIT

#### What's in this PR?

This PR add tabs to the list view and filter types in form by the tag.

#### TODO

- [x] Type dependent permissions


luca-rath:
The failing travis builds have nothing to do with this pull request

Admin/ArticleAdmin.php
```diff
@@ -114,134 +133,177 @@ function(Localization $localization) {
...
+                $viewCollection->add(
+                    $this->viewBuilderFactory->createListViewBuilder(static::LIST_VIEW . '_' . $typeKey, '/:locale/' . $typeKey)
+                        ->setResourceKey('articles')
+                        ->setListKey('articles')
+                        ->setTabTitle($typeConfig['title'])
+                        ->addListAdapters(['table'])
+                        ->addLocales($locales)
+                        ->addRequestParameters(['types' => $typeKey])
+                        ->setDefaultLocale($locales[0])
+                        ->setAddView(static::ADD_FORM_VIEW . '_' . $typeKey)
```
luca-rath, line 232: @wachterjohannes should the `AddView` and `EditView` only be available if the user has `edit` permission for this article type?

wachterjohannes: @luca-rath how is it in the contact-bundle for example? we should reuse the behaviour from there.

luca-rath: @wachterjohannes Without `edit` permission, no view is added at all. But then we should drop the `live` permission for article types, because it's only for displaying the navigation item.

wachterjohannes: 👍

PR #422, opened 2019-12-13
Remove pageTreeRoute functionality

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes #issuenum
| Related issues/PRs | https://github.com/sulu/sulu/pull/4939
| License | MIT

#### What's in this PR?

Removed pageTreeRoute functionality

#### Why?

Because it has been moved to sulu core

PR #450, opened 2020-03-30
Add missing response and requestFormat variable

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add missing variables for proper functionality.

### DIFF
--- a/Controller/WebsiteArticleController.php
+++ b/Controller/WebsiteArticleController.php
@@ -61,7 +61,7 @@ class WebsiteArticleController extends Controller
 
         try {
             if ($partial) {
-                $this->createResponse($request);
+                $response = $this->createResponse($request);
                 $response->setContent(
                     $this->renderBlock(
                         $viewTemplate,
@@ -145,7 +145,7 @@ class WebsiteArticleController extends Controller
         // we need to set the content type ourselves here
         // else symfony will use the accept header of the client and the page could be cached with false content-type
         // see following symfony issue: https://github.com/symfony/symfony/issues/35694
-        $mimeType = $request->getMimeType($requestFormat);
+        $mimeType = $request->getMimeType($request->getRequestFormat());
 
         if ($mimeType) {
             $response->headers->set('Content-Type', $mimeType);
--- a/Document/Behavior/RoutablePageBehavior.php
+++ b/Document/Behavior/RoutablePageBehavior.php
@@ -11,47 +11,11 @@
 
 namespace Sulu\Bundle\ArticleBundle\Document\Behavior;
 
-use Sulu\Bundle\RouteBundle\Model\RoutableInterface;
-use Sulu\Component\Content\Document\Behavior\StructureBehavior;
-use Sulu\Component\DocumentManager\Behavior\Mapping\LocaleBehavior;
-use Sulu\Component\DocumentManager\Behavior\Mapping\UuidBehavior;
-use Sulu\Component\Route\Document\Behavior\RoutableBehavior;
+use Sulu\Component\Route\Document\Behavior\RoutableBehavior as BaseRoutableBehavior;
 
 /**
  * This behavior has to be attached to documents which should have a sulu-route but managed by their parent.
  */
-interface RoutablePageBehavior extends RoutableInterface, UuidBehavior, LocaleBehavior, StructureBehavior, RoutableBehavior
+interface RoutablePageBehavior extends BaseRoutableBehavior
 {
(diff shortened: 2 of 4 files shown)


Controller/WebsiteArticleController.php
```diff
@@ -61,7 +61,7 @@ protected function renderArticle(Request $request, ArticleInterface $object, $vi
 
         try {
             if ($partial) {
-                $this->createResponse($request);
+                $response = $this->createResponse($request);
```
nnatter: we should really enable phpstan in this repository..

Document/Behavior/RoutablePageBehavior.php
```diff
@@ -15,11 +15,12 @@
 use Sulu\Component\Content\Document\Behavior\StructureBehavior;
 use Sulu\Component\DocumentManager\Behavior\Mapping\LocaleBehavior;
 use Sulu\Component\DocumentManager\Behavior\Mapping\UuidBehavior;
+use Sulu\Component\Route\Document\Behavior\RoutableBehavior;
 
 /**
  * This behavior has to be attached to documents which should have a sulu-route but managed by their parent.
  */
-interface RoutablePageBehavior extends RoutableInterface, UuidBehavior, LocaleBehavior, StructureBehavior
+interface RoutablePageBehavior extends RoutableInterface, UuidBehavior, LocaleBehavior, StructureBehavior, RoutableBehavior
```
alexander-schranz: @Prokyonn This needs a higher min requirement of sulu/sulu since when does `RoutableBehavior` exist ?

alexander-schranz: @wachterjohannes @danrot this sounds like a bc break in sulu/sulu that it nows errors and didn't error before.

alexander-schranz: Oh okay this is a page tree route interface which we had not yet implemented in 2.0 before.

alexander-schranz: Think #422 is the missing part here @luca-rath?

luca-rath: Yes, this change will be introduced when #422 is merged. This should not be needed here i guess.

Document/Behavior/RoutablePageBehavior.php
```diff
@@ -11,46 +11,11 @@
...
-use Sulu\Component\Content\Document\Behavior\StructureBehavior;
-use Sulu\Component\DocumentManager\Behavior\Mapping\LocaleBehavior;
-use Sulu\Component\DocumentManager\Behavior\Mapping\UuidBehavior;
+use Sulu\Component\Route\Document\Behavior\RoutableBehavior as BaseRoutableBehavior;
 
 /**
  * This behavior has to be attached to documents which should have a sulu-route but managed by their parent.
  */
-interface RoutablePageBehavior extends RoutableInterface, UuidBehavior, LocaleBehavior, StructureBehavior
+interface RoutablePageBehavior extends BaseRoutableBehavior
```
luca-rath: @Prokyonn why did you change this here? I don't think, that this is needed

composer.json
```diff
@@ -24,7 +24,7 @@
         "jms/serializer-bundle": "^3.3",
         "ongr/elasticsearch-bundle": "^5.2",
         "ongr/elasticsearch-dsl": "~5.0 || ~6.0",
-        "sulu/sulu": "^2.0.3",
+        "sulu/sulu": "^2.0.5",
```
luca-rath: I would prefer requiring `^2.0.6`, because `2.0.5` introduced a bug and `2.0.6` is just the fix for it. WDYT?

alexander-schranz: The bug has nothing todo with article bundle so I keep this and will merge this as it is to fix the current state.

Document/Behavior/RoutablePageBehavior.php
Prokyonn: It is needed for some functional tests, but the rest of the PageTreeRoute will be fixed in your PR :)

alexander-schranz: If changes needed here this can be done in #422 but currently this is needed that the CI is running without an interface error.

alexander-schranz:
Will merge this one hope we get #422 merged soon so the CI is running again without errors.

Issue #430, opened 2020-01-19
ArticleBundle even without Elasticsearch


csaeum:
ES is currently a requirement for ArticleBundle

especially for small pages or blogs that are just starting, this is simply too big.
I would appreciate it if you could switch via config.
First a normal search using SQL Query as standard and whoever wants can then switch to ES.

TheCadien:
i write my own news bundle for exactly that case. 
its still in progress, but feel free to take a look
https://github.com/TheCadien/sulu-news-bundle

alexander-schranz:
@TheCadien we are improving adding content behaviours (Template, Seo, Excerpt) to custom entity with https://github.com/sulu/SuluContentBundle. Its still in development but you can see an example implementation in the tests folder of the bundle: https://github.com/sulu/SuluContentBundle/tree/master/Tests/Application/ExampleTestBundle
So you don't need to reinvent all by yourself.

TheCadien:
@alexander-schranz oh realy nice to hear !

csaeum:
@alexander-schranz 
Ok, everything sounds good. But to what extent could the IT be outsourced. So that the requirements with ES are not too great for small pages?

alexander-schranz:
@csaeum not sure if I understand you correctly if you want to outsource elastic search there are SaaS for example [elastic cloud](https://www.elastic.co/de/elasticsearch/service/pricing). But never used something like that.

csaeum:
It is clear to me that I can outsource IT, but for small websites or projects IT itself is too much. A normal search would suffice for maybe 50 subpages.

My question is why you include it in the Article Bundle as an obligation and not outsource it as a bundle!

So small pages can also operate the ArticleBundle without ES and have a search that relies on SQL.

If the project then becomes larger, the operator of the site can decide whether to install ES himself or get a SAAS from ES and install an ES bundle from SULU (which may then be available) and enter this in the config from now on no longer the SQL search is taken but the ES search.

Exactly for this reason @TheCadien created his bundle because the ES is just too big for his small projects.

markuman:
Elasticsearch is expensive - in any case.  
Self-driven costs a lot of ressources (cpu, memory, time (people, knowledge)).  
Elasticsearch SaaS solutions are very very expensive (money) too. 

So outsourcing is **not** a solution!

This bundle is simply bad designed. 
Have you ever consider to use just the `fulltext index` feature of the existing rdbms (mysql/mariadb/postgres)?

[...]

PR #459, opened 2020-04-16
Adjust README.md structure and content

nnatter:
See: https://github.com/nnatter/SuluArticleBundle/blob/enhancement/readme-structure/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,11 +63,10 @@ find out **how to set up and configure the SuluArticleBundle** to your specific 
 
 ### Article Characteristics
 
-Similar to Sulu pages, articles are **configured via templates** and can include additional **SEO and excerpt information**.
+Like Sulu pages, articles are **configured via templates** and can include additional **SEO and excerpt information**.
 Moreover, articles support the same **drafting, publishing and versioning functionality** as provided by pages.
 In contrast to Sulu pages, articles are **managed in a flat list** instead of a tree structure. Furthermore, unlike 
-the Sulu page tree, the article bundle is **optimized for managing a big number of articles** without neglecting 
-performance.
+the Sulu page tree, the article bundle is **optimized for managing a big number of articles**.
 
 
 ### Elasticsearch dependency
--- a/README.md
+++ b/README.md
@@ -28,11 +28,11 @@
 </p>
 <br/>
 
-The **SuluArticleBundle** integrates a performance optimized way for managing articles into the [Sulu](https://sulu.io/) 
+The **SuluArticleBundle** integrates a performance optimized way for managing articles in the [Sulu](https://sulu.io/) 
 content management system. In the context of this bundle, **articles are localized content-rich entities** that are 
 **manageable via the Sulu administration interface** and can be **rendered on a website delivered by Sulu**. 
-This makes them a good choice for managing things like blog posts, products or even recipes inside of a Sulu project. 
-To keep things clean, the bundle allows to **manage different types of articles via separated lists** in the 
+This makes them a good choice for managing things like blog posts, products or even recipes in a Sulu project. 
+In order to keep things clean, the bundle allows to **manage different types of articles via separated lists** in the 
 administration interface.
 
 <br/>
@@ -65,9 +65,9 @@ find out **how to set up and configure the SuluArticleBundle** to your specific 
 
 Similar to Sulu pages, articles are **configured via templates** and can include additional **SEO and excerpt information**.
 Moreover, articles support the same **drafting, publishing and versioning functionality** as provided by pages.
-On the other hand, in contrast to Sulu pages, articles are **managed in a flat list** instead of a tree structure. 
-Furthermore, unlike the Sulu page tree, the article bundle is **optimized for managing a big number of articles** without 
-any performance decline.
+In contrast to Sulu pages, articles are **managed in a flat list** instead of a tree structure. Furthermore, unlike 
+the Sulu page tree, the article bundle is **optimized for managing a big number of articles** without neglecting 
+performance.
 
 
 ### Elasticsearch dependency
--- a/README.md
+++ b/README.md
@@ -1,47 +1,95 @@
-# SuluArticleBundle
+<h1 align="center">SuluArticleBundle</h1>
 
-[![Build Status](https://travis-ci.org/sulu/SuluArticleBundle.svg?branch=master)](https://travis-ci.org/sulu/SuluArticleBundle/branches)
-[![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/sulu/SuluArticleBundle/badges/quality-score.png)](https://scrutinizer-ci.com/g/sulu/SuluArticleBundle/)
-[![Code Coverage](https://scrutinizer-ci.com/g/sulu/SuluArticleBundle/badges/coverage.png)](https://scrutinizer-ci.com/g/sulu/SuluArticleBundle/)
-[![StyleCI](https://styleci.io/repos/61883398/shield)](https://styleci.io/repos/61883398)
+<p align="center">
+    <a href="https://sulu.io/" target="_blank">
+        <img width="30%" src="https://sulu.io/uploads/media/800x/00/230-Official%20Bundle%20Seal.svg?v=2-6&inline=1" alt="Official Sulu Bundle Badge">
+    </a>
+</p>
 
-The SuluArticleBundle adds support for managing articles in Sulu. Articles can be used in a lot of different ways to
-manage unstructured data with an own URL in an admin-list. Most of the features, which can be used in pages, can also
-be used on articles - like templates, versioning, drafting, publishing and automation.
+<p align="center">
+    <a href="https://github.com/sulu/SuluArticleBundle/blob/master/LICENSE" target="_blank">
+        <img src="https://img.shields.io/github/license/sulu/SuluArticleBundle.svg" alt="GitHub license">
+    </a>
+    <a href="https://github.com/sulu/SuluArticleBundle/releases" target="_blank">
+        <img src="https://img.shields.io/github/tag/sulu/SuluArticleBundle.svg" alt="GitHub tag (latest SemVer)">
+    </a>
+    <a href="https://github.com/sulu/SuluArticleBundle/actions" target="_blank">
+        <img src="https://img.shields.io/github/workflow/status/sulu/SuluArticleBundle/Test%20application/master.svg?label=github-actions" alt="GitHub actions status">
+    </a>
+    <a href="https://travis-ci.org/sulu/SuluArticleBundle/branches" target="_blank">
+        <img src="https://img.shields.io/travis/sulu/SuluArticleBundle/master.svg?label=travis" alt="Travis build">
+    </a>
+    <a href="https://circleci.com/gh/sulu/SuluArticleBundle/tree/master" target="_blank">
+        <img src="https://img.shields.io/circleci/build/github/sulu/SuluArticleBundle/master.svg?label=circleci" alt="CircleCI build">
+    </a>
+    <a href="https://github.com/sulu/sulu/releases" target="_blank">
+        <img src="https://img.shields.io/badge/sulu%20compatibility-%3E=2.0-52b6ca.svg" alt="Sulu compatibility">
+    </a>
+</p>
+<br/>
 
-Additional features included:
+The **SuluArticleBundle** integrates a performance optimized way for managing articles into the [Sulu](https://sulu.io/) 
+content management system. In the context of this bundle, **articles are localized content-rich entities** that are 
+**manageable via the Sulu administration interface** and can be **rendered on a website delivered by Sulu**. 
+This makes them a good choice for managing things like blog posts, products or even recipes inside of a Sulu project. 
+To keep things clean, the bundle allows to **manage different types of articles via separated lists** in the 
+administration interface.
 
-* Build in view-layer with elasticsearch
-* Segmentation of article-templates (called article-types)
-* Define URL schemas per type
+<br/>
+<p align="center">
+    <img width="80%" src="https://sulu.io/uploads/media/800x@2x/05/235-ezgif.gif?v=1" alt="Sulu Slideshow">
+</p>
+<br/>
 
-## Requirements
+The SuluArticleBundle is compatible with Sulu **starting from version 2.0**. Have a look at the `require` section in 
+the [composer.json](https://github.com/sulu/SuluArticleBundle/blob/master/composer.json) to find an 
+**up-to-date list of the requirements** of the bundle.
 
-* Composer
-* PHP `^7.2`
-* Sulu `^2.0.1`
-* Elasticsearch `~5.0 || ~6.0`
 
-For detailed requirements see [composer.json](https://github.com/sulu/SuluArticleBundle/blob/master/composer.json).
+## 🚀&nbsp; Installation and Documentation
 
-## Documentation
+Execute the following [composer](https://getcomposer.org/) command to add the bundle to the dependencies of your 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,47 +1,95 @@
...
+        <img src="https://img.shields.io/circleci/build/github/sulu/SuluArticleBundle/master.svg?label=circleci" alt="CircleCI build">
+    </a>
+    <a href="https://github.com/sulu/sulu/releases" target="_blank">
+        <img src="https://img.shields.io/badge/sulu%20compatibility-%3E=2.0-52b6ca.svg" alt="Sulu compatibility">
+    </a>
+</p>
+<br/>
 
-Additional features included:
+The **SuluArticleBundle** integrates a performance optimized way for managing articles into the [Sulu](https://sulu.io/)
```
danrot, line 31: "managing article in the Sulu content management system"?

README.md
```diff
@@ -1,47 +1,95 @@
...
+        <img src="https://img.shields.io/badge/sulu%20compatibility-%3E=2.0-52b6ca.svg" alt="Sulu compatibility">
+    </a>
+</p>
+<br/>
 
-Additional features included:
+The **SuluArticleBundle** integrates a performance optimized way for managing articles into the [Sulu](https://sulu.io/) 
+content management system. In the context of this bundle, **articles are localized content-rich entities** that are 
+**manageable via the Sulu administration interface** and can be **rendered on a website delivered by Sulu**. 
+This makes them a good choice for managing things like blog posts, products or even recipes inside of a Sulu project.
```
danrot, line 34: "in a Sulu project"? "inside of" is only used if ["inside" is used as a noun](https://english.stackexchange.com/questions/61930/inside-of-vs-inside), and just "inside a Sulu project" sounds weird. I think my proposal is also easier to read, because it is more straightforward.

README.md
```diff
@@ -1,47 +1,95 @@
...
+    </a>
+</p>
+<br/>
 
-Additional features included:
+The **SuluArticleBundle** integrates a performance optimized way for managing articles into the [Sulu](https://sulu.io/) 
+content management system. In the context of this bundle, **articles are localized content-rich entities** that are 
+**manageable via the Sulu administration interface** and can be **rendered on a website delivered by Sulu**. 
+This makes them a good choice for managing things like blog posts, products or even recipes inside of a Sulu project. 
+To keep things clean, the bundle allows to **manage different types of articles via separated lists** in the
```
danrot, line 35: "In order to keep things clean, ..."

README.md
```diff
@@ -1,47 +1,95 @@
...
-[Documentation](https://github.com/sulu/SuluArticleBundle/blob/master/Resources/doc/installation.md).
 
-## License
+## 💡&nbsp; Key Concepts
 
-This bundle is under the MIT license. See the complete license [in the bundle](LICENSE)
+### Article Characteristics
 
-## Reporting an issue or a feature request
+Similar to Sulu pages, articles are **configured via templates** and can include additional **SEO and excerpt information**.
```
danrot, line 66: I would say that part is not only similar, but almost identical, isn't it?

README.md
```diff
@@ -1,47 +1,95 @@
...
+Moreover, articles support the same **drafting, publishing and versioning functionality** as provided by pages.
+On the other hand, in contrast to Sulu pages, articles are **managed in a flat list** instead of a tree structure.
```
danrot, line 68: I would only use "On the other hand" or "In contrast to Sulu pages". Probably the latter.

README.md
```diff
@@ -1,47 +1,95 @@
...
 
-This bundle is under the MIT license. See the complete license [in the bundle](LICENSE)
+### Article Characteristics
 
-## Reporting an issue or a feature request
+Similar to Sulu pages, articles are **configured via templates** and can include additional **SEO and excerpt information**.
+Moreover, articles support the same **drafting, publishing and versioning functionality** as provided by pages.
+On the other hand, in contrast to Sulu pages, articles are **managed in a flat list** instead of a tree structure. 
+Furthermore, unlike the Sulu page tree, the article bundle is **optimized for managing a big number of articles** without 
+any performance decline.
```
danrot, line 70: "any" is probably a very strong word here, isn't it? Would phrase it more carefully, maybe "without any noticable perfomance impact"? I am pretty sure that it is less performant with millions of entries than it is with 10.

README.md
```diff
@@ -1,47 +1,95 @@
...
-Issues and feature requests are tracked in the [Github issue tracker](https://github.com/Sulu/SuluArticleBundle/issues).
 
-When reporting a bug, it may be a good idea to reproduce it in a basic project built using the
-[Sulu Minimal Edition](https://github.com/sulu/sulu-minimal) to allow developers of the bundle to reproduce the issue
-by simply cloning it and following some steps.
+### Elasticsearch dependency
+
+The SuluArticleBundle was originally developed to be used in the publishing industry. To allow for **implementing 
+efficient and flexible search endpoints** for a big number of articles, the bundle **utilizes a view layer that is 
+stored in an Elasticsearch index**. This makes Elasticsearch a **hard dependency** of the bundle. If your project will
```
danrot, line 77: These flexible search endpoints could also be implement with Elasticsearch being an optional dependency. So if people don't want to use Elasticsearch this might sound like a lame excuse to them. I am talking especially about the "This makes Elasticsearch..." sentence. Would probably be more honest here. Maybe "For this reason we chose to make Elasticsearch a hard dependency"? Or be even more straight and mention that we would like to remove Elasticsearch at some point in the future? People keep asking that anyway, therefore it might be OK to mention that in the README already.

README.md
```diff
@@ -1,47 +1,95 @@
...
 
-When reporting a bug, it may be a good idea to reproduce it in a basic project built using the
-[Sulu Minimal Edition](https://github.com/sulu/sulu-minimal) to allow developers of the bundle to reproduce the issue
-by simply cloning it and following some steps.
+### Elasticsearch dependency
+
+The SuluArticleBundle was originally developed to be used in the publishing industry. To allow for **implementing 
+efficient and flexible search endpoints** for a big number of articles, the bundle **utilizes a view layer that is 
+stored in an Elasticsearch index**. This makes Elasticsearch a **hard dependency** of the bundle. If your project will 
+not utilize these performance optimizations, it might be more convenient to use Sulu pages below a specific parent
```
danrot, line 78: I think this is confusing as well, somehow related to the previous comment. I think one of the biggest advantages is viewing documents as lists, especially from a user's perspective.

composer.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "sulu/article-bundle",
-    "description": "Sulu bundle for managing articles",
+    "description": "Bundle for managing localized content-rich entities like blog-posts in the Sulu content management system.",
```
danrot, line 3: Since this is not a full sentence I would remove the full stop at the end.

README.md
nnatter: `article`?

README.md
nnatter: Would you use `identical` then? I dont know the implementation of the bundle good enough to assess this. For example, there is no such things as article types for pages.

README.md
nnatter: Yes you are right. I am not a fan of the dependency and had a bit of a hard time phrasing this in a non-negative way to be honest. But I think that it is important to mention the dependency in the readme and give at least some reasons for it. How would you phrase this section?

README.md
nnatter: Probably yes, but I think it is important to provide an alternative solutions for simple usecases that does not require Elasticsearch (see eg #430). Again, I had a hard time finding a good wording for this.. How would you phrase it?

composer.json
nnatter: Would you remove the stop for the `sulu/skeleton` and `sulu/sulu` too?

README.md
danrot: Typo, wanted to keep `articles`, this one was about the preposition.

README.md
danrot: I am not even talking about the implementation, but for the user there is no difference, especially for the things you've listed here. To me it sounds like there is no SEO and excerpt information for pages. Maybe "Articles reuse the existing functionality already available for pages, like configuration via templates and SEO and excerpt information". You are listing the differences in the subsequent paragraphs anyway.

README.md
danrot: Hm, maybe we should go for the honest way?

"Currently there is no way to use this bundle without Elasticsearch, but we are planning to remove that hard dependency in the future."

That might lead less people to ask this question in our Slack channel, because it is already answered publicly. But I think we should also take the opinions of @chirimoya and @wachterjohannes into account.

README.md
danrot: Providing an alternative solution is a good idea. WDYT about something like this:

If you can't or don't want to make Elasticsearch a dependency of your project, you can still use Sulu pages for certain use cases. However, be aware that the article list is more comfortable to use with a large amount of entities and the performance might be not as good.

composer.json
danrot: If what is written in there is not full sentences, yes :slightly_smiling_face:

README.md
chirimoya: "Based on the initial requirements regarding performance and scalability, we decided to add an Elasticsearch view-layer as hard dependency. We are aware of the additional complexity and going to make this dependency optional with the next major release." something like that, would be fine for me.

README.md
chirimoya: +1 for this hint

README.md
nnatter: Sorry, I dont really like that suggestion. IMO it sounds wrong that articles are reusing something. Also `configuration via templates` is not a functionality to me.

Regarding to `but for the user there is no difference, especially for the things you've listed here`. Isnt the article type feature exactly this? It concerns the template, it is not possible for pages and it is relevant for the user of the bundle (user in the context of the one who looks for bundles on github).

danrot: > Sorry, I am not happy with that suggestion. IMO it sounds wrong that articles are reusing something. Also `configuration via templates` is not a functionality to me.

How else would you call it? This bundle is reusing this functionality, otherwise the template parsing would have to be implemented here again, which is not the case. And for me configuring something is also a functionality.

> Regarding to `but for the user there is no difference, especially for the things you've listed here`. Isnt the article type feature exactly this? It concerns the template, it is not possible for pages and it is relevant for the user of the bundle (user in the context of the one who looks for bundles on github).

What should the article feature be? I don't really get that. This is just a tag on the XML template, isn't it? And the article types are not even mentioned in the sentence we are talking about. IMO this could be mentioned in one of the next sentences, e.g. in the one using "In contrast to Sulu pages".

README.md
```diff
@@ -1,47 +1,98 @@
...
+## 💡&nbsp; Key Concepts
 
-This bundle is under the MIT license. See the complete license [in the bundle](LICENSE)
+### Article Characteristics
 
-## Reporting an issue or a feature request
+Similar to Sulu pages, articles are **configured via templates** and can include additional **SEO and excerpt information**.
+Moreover, articles support the same **drafting, publishing and versioning functionality** as provided by pages.
+In contrast to Sulu pages, articles are **managed in a flat list** instead of a tree structure. Furthermore, unlike 
+the Sulu page tree, the article bundle is **optimized for managing a big number of articles** without neglecting
```
danrot, line 69: Can we just write that the performance might be better in the Sulu page tree? This way it sounds like we are neglecting performance in the Sulu page tree.

nnatter: I think it would be useful if you propose full sentences from now on.

README.md
nnatter: > This bundle is reusing this functionality
You wrote `article` above - that is a difference to me.

> How else would you call it
To me, it is a characteristic of articles that they are configured via templates, not a feature of them.

> What should the article feature be? 
An example why I dont think it is correct to write that the `configuration via templates` is identical to pages.

Not sure if this discussion is leading anywhere useful to be honest 🙈

danrot: > > This bundle is reusing this functionality
> 
> You wrote `article` above - that is a difference to me.

Then why not rewrite it to "This bundle reuses existing functionality, like the configuration via templates and the inclusion of SEO and excerpt information."?

> > How else would you call it
> 
> To me, it is a characteristic (or maybe feature, but not in the context of software engineering) of articles that they are configured via templates, not a functionality of them.
> > What should the article feature be?
> 
> An example why I dont think it is correct to write that the `configuration via templates` is identical to pages.

Yeah, that's the only difference AFAIK. But this bundle still reuses the functionality, even the handling of tags (which are used to define the different article types) is implemented in the Sulu core. From my POV all issues would be resolved if we say that the bundle (not the article) reuses that functionality.

README.md
danrot: WDYT about "Furthermore, unlike the Sulu page tree, the SuluArticleBundle is **optimized for managing a big number of articles** and tries to achieve a better performance."?

README.md
wachterjohannes: I would go with `Similar to Sulu pages, articles are **configured via templates** and can include additional **SEO and excerpt information**.` i think from the developer perspective it does not make any difference and it would bring up the question: "If it is the same why using it?"

wachterjohannes: additionally the key of the paragraph is to bring the concept of articles and not the functionalities of it.

danrot: The difference is explained in the next sentence, so that's not a big problem IMO. The problem with "similar" is, that whenever I read that, I immediately think "How is it different?". And in that case it is not different. And since I am asking myself that, I might be tempted to think that the difference is that it also includes SEO and excerpt information, as the second part of the sentence suggests.

For me it would also be fine if we write "Like Sulu pages, articles are configured via templates and can include additional SEO and excerpt information.". Then it is clearer to me that all that we are talking about what works the same way.

README.md
nnatter: I am sorry, I really dont want to be nitpicky, but I dont feel that this is better.. 
Firstly, `tries to achieve` to achieve does not sound very confident for something that is a key selling point 🙈 
Secondly `a better performance` does not feel like the end of a complete sentence. Better than what? (i know, better than the page tree, but that conflicts gramatically with the `unlike` that is already part of the sentence)

README.md
danrot: > additionally the key of the paragraph is to bring the concept of articles and not the functionalities of it.

I don't care if you call it characteristics, concepts, features, functionality or anything else, especially since none of these words are used in that sentence :see_no_evil: Fact is, that both of these things, the configuration and SEO and excerpt information, work just like for Sulu pages. That's why I would go with "Like Sulu pages" instead of "Similar to Sulu pages".

README.md
danrot: Well, you are nitpicky :see_no_evil: 

However, I also feel like we can remove the second part of the sentence anyway. Performance is mentioned in the next paragraph about elastic search anyway, and "optimized for managing a big number of articles" includes performance for me as well.

README.md
nnatter: > I don't care if you call it characteristics, concepts, features, functionality or anything else, especially since none of these words are used in that sentence 

But the title of the paragraph is `Article Characteristics`. The intention of this section is to give the reader a feeling about what an article is and how it compares to a page.

`Like Sulu pages` sounds good to me.

danrot: > > I don't care if you call it characteristics, concepts, features, functionality or anything else, especially since none of these words are used in that sentence
> 
> But the title of the paragraph is `Article Characteristics`. The intention of this section is to give the reader a feeling about what an article is and how it compares to a page.
> 
> `Like Sulu pages` sounds good to me.

My point is also not influenced by the title at all :stuck_out_tongue: But that is a discussion that is really useless :see_no_evil:

PR #484, opened 2020-05-27
Remove installation instructions that are not necessary anymore

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT

#### What's in this PR?

This PR adjusts the installation instructions that are not true anymore.

#### Why?

Because otherwise people receive an error while running them, which might confuse them.

### DIFF
--- a/Resources/doc/default.html.twig
+++ b/Resources/doc/default.html.twig
@@ -1,15 +1,5 @@
 {% extends "base.html.twig" %}
 
-{% block meta %}
-    {% include "SuluWebsiteBundle:Extension:seo.html.twig" with {
-        "seo": extension.seo|default([]),
-        "content": content|default([]),
-        "urls": urls|default([]),
-        "shadowBaseLocale": shadowBaseLocale|default(),
-        "defaultLocale": request.defaultLocale|default('de')
-    } %}
-{% endblock %}
-
 {% block content %}
     <h1>{{ content.title }}</h1>
 
--- a/Resources/doc/installation.md
+++ b/Resources/doc/installation.md
@@ -2,23 +2,12 @@
 
 ### ElasticSearch
 
-The SuluArticleBundle requires a running elasticsearch `^5.0`.
+The SuluArticleBundle requires a running elasticsearch `^5.0` or `^6.0`.
 
 ## Install the bundle
 
 ```bash
 composer require sulu/article-bundle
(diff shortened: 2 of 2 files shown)


Resources/doc/default.html.twig
```diff
@@ -1,15 +1,5 @@
 {% extends "base.html.twig" %}
 
-{% block meta %}
```
danrot, line 3: This seems to be included in the [base twig template](https://github.com/sulu/skeleton/blob/fa709e5/templates/base.html.twig#L9-L15), so I don't see any point in doing this here again. Additionally the include in here follows the old `:` syntax, which causes an error now in the latest version of Symfony.

Resources/doc/installation.md
```diff
@@ -2,25 +2,14 @@
...
 ```
 
-### Add bundles to AbstractKernel
-
-The bundle need to be registered after the `SuluCoreBundle` and `SuluDocumentManagerBundle`.
-
-```php
-/* config/bundles.php */
-
-Sulu\Bundle\ArticleBundle\SuluArticleBundle::class => ['all' => true],
```
danrot, line 20: These bundles are now added automatically. @alexander-schranz wanted to leave it in here anyway, but I think this installation guide should be as short as possible (also the rest of it it not very exhaustive).

And people might not read the text coming before this code block, and copy it anyway, no matter if it is already added.

alexander-schranz: Would not remove this symfony always writes the following into there docs:

> If you are not using Flex, you also have to enable the bundle by adding the following lines in the `config/bundles.php`:

danrot: But people start with the `sulu/skeleton`, so they are using it anyway? Would make it as easy as possible for those, the other ones are an edge case for me.

alexander-schranz: Still thing we should document all, most people look only in the docs when they have problems. Maybe move it to a `# Troubleshooting` section down, but would not removed it.

danrot: I still think this is a non-issue, and makes the installation guide unnecessarily complicated, but have added it anyway.

PR #494, opened 2020-06-22
Fix pagination default limit for smart content

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

The defaultLimit should not have any effect on paginated smart content it should just limit the pageSize not the whole totals.

#### Why?

Currently pagination over the default limit doesn't work and will not return all articles. So for example a smart content with `12` per page and 118 articles will just return the first 100 articles, show hasNextPage as true but will fail with OutofBoundException on Page 10.

### DIFF
--- a/Content/ArticleDataProvider.php
+++ b/Content/ArticleDataProvider.php
@@ -234,8 +234,14 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
     private function hasNextPage(\Countable $queryResult, $limit, $page, $pageSize)
     {
         $count = $queryResult->count();
-        if ($limit && $limit < $count) {
-            $count = $limit;
+
+        if (null === $pageSize || $pageSize > $this->defaultLimit) {
+            $pageSize = $this->defaultLimit;
+        }
+
+        $offset = ($page - 1) * $pageSize;
+        if ($limit && $offset + $pageSize > $limit) {
+            return false;
         }
 
         return $count > ($page * $pageSize);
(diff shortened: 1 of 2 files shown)


Content/ArticleDataProvider.php
```diff
@@ -234,8 +234,14 @@ private function getWebspaceKey(array $propertyParameter, array $options)
     private function hasNextPage(\Countable $queryResult, $limit, $page, $pageSize)
     {
         $count = $queryResult->count();
-        if ($limit && $limit < $count) {
-            $count = $limit;
+
+        if (null === $pageSize || $pageSize > $this->defaultLimit) {
```
nnatter, line 238: is the `defaultLimit` also a maximum limit? 🤔

Content/ArticleDataProvider.php
```diff
@@ -234,8 +234,14 @@ private function getWebspaceKey(array $propertyParameter, array $options)
...
+            $pageSize = $this->defaultLimit;
+        }
+
+        $offset = ($page - 1) * $pageSize;
+        if ($limit && $offset + $pageSize > $limit) {
```
nnatter, line 243: is there a reason for extracting the `$offset`? its okay for me both ways, just curious 🙂

Content/ArticleDataProvider.php
alexander-schranz: No that is the current issue it affects the limit but it should not affect it when having paginated content.

Content/ArticleDataProvider.php
alexander-schranz: Just be consistent to the other method where the same logic is used

nnatter: okay figured that you did it to be consistent with the `addPagination `

Content/ArticleDataProvider.php
nnatter: okay this is needed to be consistent with the behaviour of the `addPagination ` method 👍

PR #502, opened 2020-07-01
Order article types by configuration

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Order article types by configuration.

#### Why?

Currently its random by in the order how the operation system read the structure files.

### DIFF
--- a/Admin/ArticleAdmin.php
+++ b/Admin/ArticleAdmin.php
@@ -307,18 +307,10 @@ class ArticleAdmin extends Admin
      */
     public function getSecurityContexts()
     {
-        $types = [];
         $securityContext = [];
 
-        /** @var StructureBridge $structure */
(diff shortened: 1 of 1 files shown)


Admin/ArticleAdmin.php
```diff
@@ -347,11 +347,15 @@ private function getTypes(): array
     {
         $types = [];
 
+        foreach ($this->articleTypeConfigurations as $typeKey => $articleTypeConfiguration) {
```
nnatter, line 350: maybe add a comment like `prefill array with keys from configuration to keep order of configuration for tabs`

alexander-schranz: Thank you! Did add it.

PR #509, opened 2020-08-03
Change translation domain of article type from backend to admin

nnatter:
See #507


nnatter:
I think we should adjust this in the `2.0` and `2.1` branch too. Unfortunately, I already merged #507 into `master`.

PR #511, opened 2020-08-04
Filter articles based on current segment in ArticleDataProvider

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/sulu/sulu/pull/5425
| License | MIT

#### What's in this PR?

This PR adjusts the `ArticleDataProvider` of the bundle to respect the current segment when filtering articles. With this PR, the `ArticleDataProvider` will return an article only if it is not assigned to a segment for the current webspace or if it is assigned to the current segment for the current webspace.

#### Why?

Because the `ArticleDataProvider` should behave similar to the `PageDataProvider`.

#### ToDo

- [ ] Fix missing `segments` variable in article twig rendering

### DIFF
--- a/Content/ArticleDataProvider.php
+++ b/Content/ArticleDataProvider.php
@@ -261,6 +261,20 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
             $search->addQuery($webspaceQuery);
         }
 
+        $segmentKey = $filters['segmentKey'] ?? null;
+        if ($segmentKey && $webspaceKey) {
+            $matchingSegmentQuery = new TermQuery('excerpt.segments.assignment_key', $webspaceKey . '#' . $segmentKey);
+
+            $noSegmentQuery = new BoolQuery();
+            $noSegmentQuery->add(new TermQuery('excerpt.segments.webspace_key', $webspaceKey), BoolQuery::MUST_NOT);
+
+            $segmentQuery = new BoolQuery();
+            $segmentQuery->add($matchingSegmentQuery, BoolQuery::SHOULD);
+            $segmentQuery->add($noSegmentQuery, BoolQuery::SHOULD);
+
+            $search->addQuery($segmentQuery);
+        }
+
         return $repository->findDocuments($search);
     }
 
--- a/Content/ArticleDataProvider.php
+++ b/Content/ArticleDataProvider.php
@@ -21,6 +21,7 @@ use ProxyManager\Factory\LazyLoadingValueHolderFactory;
 use ProxyManager\Proxy\LazyLoadingInterface;
 use Sulu\Bundle\ArticleBundle\Document\ArticleDocument;
 use Sulu\Bundle\ArticleBundle\Document\ArticleViewDocumentInterface;
+use Sulu\Bundle\PageBundle\Content\Types\SegmentSelect;
 use Sulu\Bundle\WebsiteBundle\ReferenceStore\ReferenceStoreInterface;
 use Sulu\Component\Content\Compat\PropertyParameter;
 use Sulu\Component\DocumentManager\DocumentManagerInterface;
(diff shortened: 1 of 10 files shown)


nnatter:
I cannot test this together with https://github.com/sulu/sulu/pull/5425 because travis refuses to install the branch of my `sulu/sulu` fork. Furthermore, the tests of this bundle seem to hang when executed locally.

Controller/WebsiteArticleController.php
```diff
@@ -62,8 +64,12 @@ protected function renderArticle(
...
         $content = $this->resolveArticle($object, $pageNumber);
 
-        $templateAttributeResolver = $this->getTemplateAttributeResolver();
-        $data = $templateAttributeResolver->resolve(array_merge($content, $attributes));
+        $data = $this->get('sulu_website.resolver.parameter')->resolve(
+            array_merge($content, $attributes),
+            $this->get('sulu_core.webspace.request_analyzer'),
+            null,
+            $preview
+        );
```
danrot, line 72: @alexander-schranz @wachterjohannes Does anything speak against that change? This way the available variables would be more consistent with what you get in pages.

wachterjohannes: @danrot does it change anything in the twig? if no then its ok for me

Document/CategoryViewObject.php
```diff
@@ -46,15 +46,7 @@ class CategoryViewObject
...
-     * @Property(
-     *     type="text",
-     *     options={
-     *        "fields"={
-     *            "raw"={"type"="keyword"},
-     *            "value"={"type"="text"}
-     *        }
-     *    }
-     * )
+     * @Property(type="keyword")
```
wachterjohannes, line 49: this is a BC break and should at least be documented to drop and recreate the index

Document/ExcerptViewObject.php
```diff
@@ -82,6 +82,13 @@ class ExcerptViewObject
      */
     public $tags;
 
+    /**
+     * @var SegmentViewObject[]|Collection
+     *
+     * @Embedded(class="Sulu\Bundle\ArticleBundle\Document\SegmentViewObject", multiple=true)
+     */
+    public $segments;
```
wachterjohannes, line 90: ah i see this change also need a drop and recreate - so the change before is OK but although it has to be documented in the UPGRADE.md file

Controller/WebsiteArticleController.php
danrot: This is the diff between the two output:

```diff
1c1
< array (size=34)
---
> array (size=35)
60a61,66
>   'localizations' => 
>     array (size=1)
>       'en' => 
>         array (size=2)
>           'locale' => string 'en' (length=2)
>           'url' => string 'http://127.0.0.1:8001/' (length=22)
89c95,104
<   'shadowBaseLocale' => null
---
>   'segments' => 
>     array (size=2)
>       'w' => 
>         array (size=2)
>           'title' => string 'Winter' (length=6)
>           'url' => string '/_sulu_segment_switch?segment=w&url=http://127.0.0.1:8001/articles/test' (length=71)
>       's' => 
>         array (size=2)
>           'title' => string 'Summer' (length=6)
>           'url' => string '/_sulu_segment_switch?segment=s&url=http://127.0.0.1:8001/articles/test' (length=71)
```

The `shadowBaseLocale` is gone this way, but it has always been `null` previously anyway (even on shadow pages).

danrot:
The tests should go green as soon as sulu/sulu#5425 is merged.

Content/ArticleDataProvider.php
```diff
@@ -261,6 +261,20 @@ private function getSearchResult(array $filters, ?int $limit, int $page, ?int $p
             $search->addQuery($webspaceQuery);
         }
 
+        $segmentKey = $filters['segmentKey'] ?? null;
+        if ($segmentKey && $webspaceKey) {
+            $matchingSegmentQuery = new TermQuery('excerpt.segments.assignment_key', $webspaceKey . '#' . $segmentKey);
```
nnatter, line 266: should we extract a `SEPERATOR` constant like in https://github.com/sulu/sulu/pull/5425/files#diff-9b4f014324c1c4a24f27e8397a33778bR20 ?

danrot:
@nnatter Managed to get the tests running for `--prefer-lowest` as well.

Tests/Functional/Controller/ArticleControllerTest.php
```diff
@@ -657,8 +659,10 @@ public function testPutExtensions(
         $this->assertEquals($title, $response['title']);
         $this->assertEquals($extensions['seo'], $response['ext']['seo']);
 
-        // segment is only returned for sulu versions >= 2.2
-        unset($response['ext']['excerpt']['segment']);
+        if (!class_exists(SegmentSelect::class)) {
```
nnatter, line 662: 👍

PR #529, opened 2020-11-26
Add SmartContent template filtering

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds SmartContent template filtering for articles.

#### BC Breaks/Deprecations

Describe BC breaks/deprecations here. (remove this section if not needed)

#### To Do

- [ ] Add documentation page

### DIFF
--- a/Content/ArticleDataProvider.php
+++ b/Content/ArticleDataProvider.php
@@ -75,11 +75,6 @@ class ArticleDataProvider implements DataProviderInterface, DataProviderAliasInt
     protected $defaultLimit;
 
     /**
-     * @var bool
-     */
-    private $hasAudienceTargeting;
-
-    /**
      * @var FormMetadataProvider|null
      */
     private $formMetadataProvider;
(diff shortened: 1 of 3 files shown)


Content/ArticleDataProvider.php
```diff
@@ -140,8 +164,8 @@ public function resolveDataItems(
         $page = 1,
         $pageSize = null
     ) {
+        $filters['structureTypes'] = array_merge($filters['types'] ?? [], $this->getStructureTypesProperty($propertyParameter));
```
nnatter, line 167: the mixup between `structureTypes` and `types` is confusing 😕  but i guess there is no better solution?

nnatter: maybe we can at least add a comment that tries to explain what is what?

Prokyonn: Yeah it is, we definitely should change this, but I don't think there is a better solution for the time being without changing too much

wachterjohannes: @Prokyonn yeah please add a comment

wachterjohannes (approved):

PR #533, opened 2020-12-01
Increase scrutinizer timeout

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Increase scrutinizer timeout

#### Why?

Because the github actions pipelines need longer than the previously configured timeout of 10 minutes


luca-rath:
I have no idea why scrutinizer fails ... The timeout is more than high enough, tests are passing, coverage report upload quits with `Uploading code coverage for repository "g/sulu/SuluArticleBundle" and revision "84d94a061c6e61518e34cf9393a55ff404747813"... Done`, like it should be. But scrutinizer never recieves the coverage report ... @wachterjohannes do you have an idea?

wachterjohannes:
@luca-rath i am not sure but we used the flag `--coverage-clover=coverage.clover` before. instead of `--coverage-clover=clover.xml` maybe scrutinizer is waiting for a file with this name (see docs https://scrutinizer-ci.com/docs/tools/external-code-coverage/#php)

are you sure that xdebug or pcov is installed? maybe there is no file generated

nnatter (approved):

PR #541, opened 2021-01-08
Expand composer lint script and add php-cs-fixer as dev-dependency

niklasnatter:


niklasnatter:
@luca-rath @alexander-schranz What do you think? Do you see any problems with this?

luca-rath (approved):
Looks good to me, the only thing that comes to my mind is, that `lint-container` should be executed as often as possible (best case would be on every matrix job) ... But in this specific case, if i remember correctly, it can't be executed with `--prefer-lowest`, because it was implemented in symfony 4.4 ... Just wanted to note that to keep that in mind in the future

niklasnatter:
> But in this specific case, if i remember correctly, it can't be executed with --prefer-lowest, because it was implemented in symfony 4.4

Yes, unfortunately the `lint:container` command is only available in Symfony 4.4 😕

luca-rath:
Do we still need circleci here? If nobody disagrees I would just remove that

niklasnatter:
I think circleci is the only place where we set the `ARTICLE_TEST_CASE` variable at the moment. But would be okay for me to change that in another PR 🙂

alexander-schranz:
@nnatter would not set `ARTICLE_TEST_CASE` in this testcases aslong as we do not migrate circleci tests to github actions.

niklasnatter:
> @nnatter would not set ARTICLE_TEST_CASE in this testcases aslong as we do not migrate circleci tests to github actions.

Yes, as said above - would be okay for me to migrate that in the future, but I dont think we need to do it now 🙂

alexander-schranz (approved):

alexander-schranz:
Looks fine from my side.

PR #515, opened 2020-09-09
Add annotation reader to dynamically add the document annotation

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #issuenum
| Related issues/PRs | requires https://github.com/handcraftedinthealps/ElasticsearchBundle/pull/12
| License | MIT

#### What's in this PR?

Add flag to enable/disable mapping prepend of the ongr bundle config.

#### Why?

When you add your own ViewDocument both "types" will be mapped and that ends up in an exception from elasticsearch.

PR #564, opened 2021-05-11
Update ArticleViewDocument.php

### LINKED ISSUE #563: ArticleViewDocument class cannot be parsed as document because @Document annotation is missing
https://github.com/sulu/SuluArticleBundle/blob/c7016d09074f4b57ff7f2cf371f0d49d43b4db86/Document/ArticleViewDocument.php#L23

Uncaught PHP Exception ONGR\ElasticsearchBundle\Exception\MissingDocumentAnnotationException: "Sulu\Bundle\ArticleBundle\Document\ArticleViewDocument" class cannot be parsed as document because @Document annotation is missing."

I am following your documentation at https://github.com/sulu/SuluArticleBundle/blob/2.x/Resources/doc/article-view-document.md to extend the index with some fields.

But then I get the error above. If I add an @ everything works fine again. Typo? Am I missing something?

kfrohwein:
Fixes #563

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #563
| License | MIT

#### What's in this PR?

Seems there is an @ missing or it won't be recognized as annotation.

#### Why?

I want to extend the articles index

### DIFF
--- a/Document/ArticleViewDocument.php
+++ b/Document/ArticleViewDocument.php
@@ -20,7 +20,7 @@ use ONGR\ElasticsearchBundle\Annotation\Property;
  * Indexable document for articles.
  *
  * Following annotation will be set by the annotation reader if this document is used for mapping.
- * Document(type="article")
+ * @Document(type="article")
  */
 class ArticleViewDocument implements ArticleViewDocumentInterface
 {


niklasnatter:
Hey, 
as explained in the comment above the modified line, the `@` is omitted intentionally because the annotation is set automatically by the `AnnotationReader` that was implemented in #515. Lets discuss why this does not work for you in #563

PR #576, opened 2021-05-28
Remove Sulu < 2.3 bridge

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Removes the BC-bridge, which is not necessary anymore.


Admin/ArticleAdmin.php
```diff
@@ -27,7 +27,6 @@
 use Sulu\Bundle\PageBundle\Document\BasePageDocument;
```
niklasnatter, line 27: i think we can remove the following bc layers too:
https://github.com/sulu/SuluArticleBundle/blob/b725301020e45c129ab1ae93f2fb24eaf496c039/Tests/Application/Kernel.php#L98-L105
https://github.com/sulu/SuluArticleBundle/blob/b725301020e45c129ab1ae93f2fb24eaf496c039/Tests/Application/Testing/ArticleBundleKernelBrowser.php#L17-L24

niklasnatter (approved):

PR #578, opened 2021-05-31
Use AutomationViewBuilderFactory instead of AutomationViewBuilder

niklasnatter:
Related: https://github.com/sulu/SuluAutomationBundle/pull/53

### DIFF
--- a/Admin/ArticleAdmin.php
+++ b/Admin/ArticleAdmin.php
@@ -331,11 +331,13 @@ class ArticleAdmin extends Admin
             );
 
             if (isset($this->kernelBundles['SuluAutomationBundle'])) {
-                $viewCollection->add(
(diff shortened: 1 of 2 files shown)


Admin/ArticleAdmin.php
```diff
@@ -100,14 +101,20 @@ class ArticleAdmin extends Admin
...
     public function __construct(
         ViewBuilderFactoryInterface $viewBuilderFactory,
         SecurityCheckerInterface $securityChecker,
         LocalizationManagerInterface $localizationManager,
         StructureManagerInterface $structureManager,
         array $kernelBundles,
         array $articleTypeConfigurations,
-        bool $versioningEnabled
+        bool $versioningEnabled,
+        ?AutomationViewBuilderFactoryInterface $automationViewBuilderFactory
```
alexander-schranz: This avoids an error which would need to hard remove the var/cache directory:

```suggestion
        ?AutomationViewBuilderFactoryInterface $automationViewBuilderFactory = null
```

Issue #580, opened 2021-06-14
URL cannot be refreshed automatically


speex404:
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| SuluArticleBundle Version | 2.3
| Sulu Version | 2.3
#### Actual Behavior
When changing the title, the button to refresh the url remains grayed out.
<br>

#### Expected Behavior

When changing the title, the button should become active and when selecting it, the url should be refreshed.
<br>

![image](https://user-images.githubusercontent.com/29870451/121925008-c3b3bb80-cd3c-11eb-901d-31b4b06fe661.png)

niklasnatter:
Hey,
thanks for reaching out. Refreshing an URL is only possible when using live generation (see https://github.com/sulu/SuluArticleBundle/issues/514#issuecomment-817662765).

At the moment, the default configuration of the bundle generates the URL based on the `ArticleDocument` instance after save. This allows to include things like the `id` of an article in the generated URL, but it prevents live generation. If you do not need data that is only available after save, you can change your configuration like described in https://github.com/sulu/SuluArticleBundle/issues/514#issuecomment-817662765.

niklasnatter:
Thanks for creating this! We adjusted the default configuration to allow for refreshing in #581 🙂

Issue #514, opened 2020-09-04
Support live generation of resource-locator when creating article


thenetexperts:
| Q | A
| --- | ---
| Bug? | maybe
| New Feature? | maybe
| SuluArticleBundle Version | 2.1.0
| Sulu Version | 2.2@dev

#### Actual Behavior

After installing fresh SuluArticleBundle and creating a new article, the ResourceLocator (path) is not being prefilled after adding the title. This is different behaviour learned from SuluPageBundle.
When the ResourceLocator is not being filled manually, the normal path creation will take place on article save, and thus visible in the UI when edition the created article again.

#### Expected Behavior

When entering the article's title the ResourceLocator should fill the field, making it easier to commit to the proposal or change the path.

#### Steps to Reproduce

Install SuluArticleBundle, create a new default article, fill the article's title.

#### Possible Solutions

Have not checked if this is to be fixed on admin ui side or SuluArticleBundle.

nnatter:
Hey, thanks for reporting this! I think it would be nice to make this consistent with the behaviour inside of pages too. 

Unfortunately, I think that the routing configuration of the bundle (see [Routing section](https://github.com/sulu/SuluArticleBundle/blob/master/Resources/doc/routing.md) of the docs) is not compatible with this behaviour at the moment. The routing configuration allows to access the article object via `object`. This is not possible until the article is saved 😕

niklasnatter:
Just for completion: ~This bundle does not support **live**-generation of the resource-locator (live update of the resource-locator when the user changes the title in the form) at the moment~ (No longer true, see comment below). But it does support automatic generation of the resource locator when the article is created.

If the resource locator input has no value when the article form is saved, the resource locator will be generated using the [configured route schema](https://github.com/sulu/SuluArticleBundle/blob/master/Resources/doc/routing.md). If you always want to generate the resource locator using the configured schema initially, you can hide the input field on the form when adding an article by using a `visibleCondition`:

```xml
<property name="routePath" type="route" visibleCondition="!!id">
    <meta>
        <title lang="en">Resourcelocator</title>
        <title lang="de">Adresse</title>
    </meta>

    <tag name="sulu_article.article_route"/>
</property>
```

niklasnatter:
[...]

Issue #538, opened 2020-12-18
Add copy article functionality


alexandersch:
I would like to request a copy article feature. Would be a nice addition to the article bundle.

rogamoore:
Hello everybody!

We would also be interested to have this feature and are happy to help making it happen.

We noticed that there is already some logic in the current version for copying articles so it looks like this feature was already there in an older version of the bundle:
see https://github.com/sulu/SuluArticleBundle/blob/2.x/Controller/ArticleController.php#L565

### PoC
With some hacks it's pretty easy to get it working:
- Add a toolbar action (using the same action as for the copy form feature) - see https://github.com/sulu/SuluArticleBundle/pull/610/files#diff-b70d69ce3c1966544518ae0ffa0b76390ad94b390ac0f749faceaaf71992841aR242
- The copy action expects a locale which the existing toolbar action does not provide, so for testing we need to hardcode it, see https://github.com/sulu/SuluArticleBundle/pull/610/files#diff-6bb92e406267424def35c1c8b3b9d0bcbc4e784b7dd15300af0dab93a7b94477R521
- Now we already have a copy action available:
![image](https://user-images.githubusercontent.com/5249624/167885190-de78edad-dc7d-46d2-a519-d2ad995592fa.png)
- However, it still fails in two places:
    - https://github.com/sulu/sulu/blob/2.5/src/Sulu/Component/Content/Document/Subscriber/WebspaceSubscriber.php#L93
    Here it fails in `deleteUnavailableLocales` when calling `$webspace->getAllLocalizations()`
![image](https://user-images.githubusercontent.com/5249624/167886095-11bb6021-fe58-499f-b66d-6d7878c13364.png)
- https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/RouteBundle/Document/Subscriber/RoutableSubscriber.php#L263
    Here it fails in `handleCopy` when calling the `chainRouteGenerator`
- Skipping both problematic places by adding a return statement to these methods makes the feature work and creates a draft copy of an existing article.
![image](https://user-images.githubusercontent.com/5249624/167888582-b2437376-48f6-46cd-afb1-aa2a6019930d.png)


### Next steps?
- Extend the `CopyToolbarAction` that was implemented in https://github.com/sulu/sulu/pull/6475 to provide the locale when triggering the `copy` action in the controller 
[...]

PR #581, opened 2021-06-15
Adjust documentation to configure live-generation of URL

niklasnatter:
#### What's in this PR?

This PR adjust the documentation in the repository to allow for liver-generation of the URL inside of the article template.

#### Why?

Because the current behaviour provides a bad UX and is inconsistent to page. See #514 and #580

### DIFF
--- a/Resources/doc/article-types.md
+++ b/Resources/doc/article-types.md
@@ -41,19 +41,3 @@ The type can also be used to predefine a filter for the smart-content provider f
 ```
 
 This will filter the smart-content for the given two types of articles.
-
-## Route Schema
-
-The bundle provides also a way to define a `route_schema` foreach article-type.
-
-```yml
-sulu_route:
-    mappings:
-        Sulu\Bundle\ArticleBundle\Document\ArticleDocument:
-            generator: type
-            options:
-                my_custom_type: "/custom/{object.getTitle()}"
-                my_second_custom_type: "/custom2/{object.getTitle()}"
-```
-
-See more information in [Routing](routing.md#route-schema).
--- a/Resources/doc/installation.md
+++ b/Resources/doc/installation.md
@@ -29,12 +29,7 @@ sulu_route:
         Sulu\Bundle\ArticleBundle\Document\ArticleDocument:
             generator: schema
             options:
-                route_schema: '/articles/{object.getTitle()}'
-        Sulu\Bundle\ArticleBundle\Document\ArticlePageDocument:
-            generator: article_page
-            options:
-                route_schema: '{translator.trans("page")}-{object.getPageNumber()}'
-                parent: '{object.getParent().getRoutePath()}'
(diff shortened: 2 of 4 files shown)


Resources/doc/article-types.md
```diff
@@ -41,19 +41,3 @@ The type can also be used to predefine a filter for the smart-content provider f
 ```
 
 This will filter the smart-content for the given two types of articles.
-
-## Route Schema
-
-The bundle provides also a way to define a `route_schema` foreach article-type.
```
niklasnatter, line 47: this will only work if the template does not contain any `sulu.rlp.part` tag. as an alternative to this, the template can define a specific schema likes this (https://github.com/sulu/sulu/pull/5965):

```
<property name="routePath" type="route" mandatory="true">
    <meta>
        <title>sulu_admin.url</title>
    </meta>

    <params>
        <param name="route_schema" value="/template-prefix/{implode('-', object)}"/>
    </params>
</property>
```

Resources/doc/installation.md
```diff
@@ -29,12 +29,7 @@ sulu_route:
         Sulu\Bundle\ArticleBundle\Document\ArticleDocument:
             generator: schema
             options:
-                route_schema: '/articles/{object.getTitle()}'
-        Sulu\Bundle\ArticleBundle\Document\ArticlePageDocument:
```
niklasnatter, line 33: The current version of the bundle does not support article pages. I think its a bit confusing to still include them in the documentation.

Resources/doc/article-types.md
wachterjohannes: @niklasnatter then we should add that to the docs

wachterjohannes (changes requested):
ELSE ACK

Resources/doc/article-types.md
niklasnatter: you are right - done 🙂

wachterjohannes (approved):

rogamoore:
Hi @niklasnatter,
while working on https://github.com/sulu/SuluArticleBundle/issues/538 I was wondering if when using the possibility to configure the `route_schema` directly in the XML template as described in this PR (see https://github.com/sulu/SuluArticleBundle/blob/2.x/Resources/doc/routing.md#overwrite-route-schema-in-template) if the YAML configuration (see https://github.com/sulu/SuluArticleBundle/blob/2.x/Resources/doc/routing.md#route-schema) is still needed?

niklasnatter:
I think the `route_schema` of the template will only be used for live generation when creating an article via the admin UI. If you create an article programatically (eg. via fixtures), only the configuration in the `.yaml` file will be used.

rogamoore:
Thanks! So the right way to achieve correct route generation programatically would be to duplicate the configuration also in the YAML file?
Is there any way to reuse the XML config?

PR #589, opened 2021-08-02
Fix href lang meta tag

### LINKED ISSUE #587: hreflang tag generation broken for unpublished articles
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| SuluArticleBundle Version | 2.3.1
| Sulu Version | 2.3.0

#### Actual Behavior

In article pages an unpublished page for a localization causes the other localizations to generate hreflang tags linking to that unpuplished page. Only if the article for the localization is deleted, the hreflang tag also gets deleted. If an article is created for the first time and no other localization version for the article exists, the hreflang tags are correct.   

#### Expected Behavior

Just like in normal pages, articles should not be generating hreflang tags for unpublished articles.

#### Steps to Reproduce

create an article and a version of the same article in another language. then unpublish one of them and check the hreflang tags on the other one. it should be removed and not linking to a 404 page of the unpublished articles.

#### Possible Solutions

Fix the "alternate" in "localizations" property from true to false when unpublishing an article.

currently its like this:
Bildschirmfoto 2021-07-22 um 16 45 39

should be like this:
Bildschirmfoto 2021-07-22 um 16 45 10

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | fixes #587
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Fix href lang meta tag

#### BC Breaks/Deprecations

Instantiating the `WebsiteArticleUrlsSubscriber` without the `$articleRouteDefaultsProvider` argument is deprecated

### DIFF
--- a/Document/Serializer/WebsiteArticleUrlsSubscriber.php
+++ b/Document/Serializer/WebsiteArticleUrlsSubscriber.php
@@ -17,7 +17,9 @@ use JMS\Serializer\EventDispatcher\ObjectEvent;
 use JMS\Serializer\Metadata\StaticPropertyMetadata;
 use JMS\Serializer\Visitor\SerializationVisitorInterface;
 use Sulu\Bundle\ArticleBundle\Document\ArticleDocument;
+use Sulu\Bundle\ArticleBundle\Routing\ArticleRouteDefaultProvider;
 use Sulu\Bundle\RouteBundle\Entity\RouteRepositoryInterface;
+use Sulu\Bundle\RouteBundle\Model\RouteInterface;
 use Sulu\Component\Webspace\Analyzer\Attributes\RequestAttributes;
 use Sulu\Component\Webspace\Manager\WebspaceManagerInterface;
 use Sulu\Component\Webspace\Webspace;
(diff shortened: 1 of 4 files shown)


Document/Serializer/WebsiteArticleUrlsSubscriber.php
```diff
@@ -99,18 +115,35 @@ public function addUrlsOnPostSerialize(ObjectEvent $event): void
...
+    {
+        if (!$route) {
+            return false;
+        }
+
+        if (null === $this->articleRouteDefaultsProvider) {
+            return true;
+        }
+
+        return $this->articleRouteDefaultsProvider->isPublished(
```
alexander-schranz: `$this->articleRouteDefaultsProvider->isPublished` is loading the whole document in the given localization. This is something we can not do for every language.

I think we maybe should correctly unpublish the `route` so it does not exist like it is when the article never was published or how is that currently  handled? And what does currently happen for pages when they are unpublished is the route there also removed?

luca-rath: The `Route` entity doesn't have information about if a route is published or not. We would have to delete the route on unpublish

luca-rath: No, the route is not being removed after unpublishing a page

luca-rath: The alternate tags are generated based on the following: https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/DocumentManagerBundle/Bridge/DocumentInspector.php#L300

luca-rath: Also, if we would remove the article routes on unpublish, we would loose all history routes

luca-rath: But introducing a `published` state on the route itself wouldn't be that bad i guess

luca-rath: @alexander-schranz How should we proceed here?

alexander-schranz (approved):

PR #610, opened 2022-05-11
Add toolbar action to copy article

### LINKED ISSUE #538: Add copy article functionality

rogamoore:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #538 
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

This adds the possibility to copy an existing article. It's similar to the "Copy Form" (https://github.com/sulu/SuluFormBundle/pull/316) and "Copy Snippet" (https://github.com/sulu/sulu/pull/6859) features.

Needs this change in `sulu/sulu`: https://github.com/sulu/sulu/pull/6859/files#diff-579b061b25841f1a028b4adc7b447e987e298306a441dd13346d08de3ec2d5ba, which is currently only in 2.6 - maybe it can be backported to allow the feature to work with Sulu >=2.4.

Please see also #538 for related discussion.

### DIFF
--- a/Admin/ArticleAdmin.php
+++ b/Admin/ArticleAdmin.php
@@ -228,13 +228,21 @@ class ArticleAdmin extends Admin
 
             if ($this->securityChecker->hasPermission(static::SECURITY_CONTEXT, PermissionTypes::LIVE)
                 && $this->securityChecker->hasPermission(static::getArticleSecurityContext($typeKey), PermissionTypes::LIVE)) {
+
+                $editDropdownToolbarActions = [
+                    new ToolbarAction('sulu_admin.delete_draft'),
+                    new ToolbarAction('sulu_admin.set_unpublished'),
+                    new ToolbarAction('sulu_admin.copy'),
+                ];
+
+                if (\count($locales) > 1) {
+                    $editDropdownToolbarActions[] = new ToolbarAction('sulu_admin.copy_locale');
+                }
+
                 $formToolbarActionsWithType[] = new DropdownToolbarAction(
                     'sulu_admin.edit',
                     'su-pen',
-                    [
-                        new ToolbarAction('sulu_admin.delete_draft'),
-                        new ToolbarAction('sulu_admin.set_unpublished'),
-                    ]
+                    $editDropdownToolbarActions
                 );
             }
 
--- a/Admin/ArticleAdmin.php
+++ b/Admin/ArticleAdmin.php
@@ -228,7 +228,6 @@ class ArticleAdmin extends Admin
 
             if ($this->securityChecker->hasPermission(static::SECURITY_CONTEXT, PermissionTypes::LIVE)
                 && $this->securityChecker->hasPermission(static::getArticleSecurityContext($typeKey), PermissionTypes::LIVE)) {
-
                 $editDropdownToolbarActions = [
                     new ToolbarAction('sulu_admin.delete_draft'),
                     new ToolbarAction('sulu_admin.set_unpublished'),
--- a/Controller/ArticleController.php
+++ b/Controller/ArticleController.php
@@ -528,7 +528,7 @@ class ArticleController extends AbstractRestController implements ClassResourceI
     {
         // extract parameter
         $action = $this->getRequestParameter($request, 'action', true);
-        $locale = $this->getRequestParameter($request, 'locale', true);
+        $locale = $this->getRequestParameter($request, 'locale', false);
 
         // prepare vars
         $view = null;
--- a/Tests/Functional/Controller/ArticleControllerTest.php
+++ b/Tests/Functional/Controller/ArticleControllerTest.php
@@ -1193,6 +1193,28 @@ class ArticleControllerTest extends SuluTestCase
         $this->assertNull($this->findViewDocument($article2['id'], 'de'));
     }
 
+    public function testCopy()
+    {
+        // create article
+        $sourceArticle = $this->testPost('Sulu ist toll - Artikel 1');
+        $this->publish($sourceArticle['id']);
+        $this->flush();
+
+        // trigger copy action
+        $this->client->jsonRequest('POST',
+            \sprintf('/api/articles/%s?action=copy', $sourceArticle['id'])
+        );
+        $response = $this->client->getResponse();
+
+        $result = \json_decode($response->getContent(), true);
+        $this->assertHttpStatusCode(200, $response);
+
+        $this->assertNotEquals($sourceArticle['id'], $result['id']);
+        $this->assertEquals($sourceArticle['title'], $result['title']);
+        $this->assertEquals($sourceArticle['articleType'], $result['articleType']);
+        $this->assertEquals(false, $result['publishedState']);
+    }
+
     public function testCopyLocaleWithoutSource()
     {
         // prepare vars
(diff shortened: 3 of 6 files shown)


rogamoore:
@alexander-schranz PR has been updated, I also added a basic test.

Admin/ArticleAdmin.php
```diff
@@ -228,13 +228,21 @@ function(Localization $localization) {
...
+                    new ToolbarAction('sulu_admin.delete_draft'),
+                    new ToolbarAction('sulu_admin.set_unpublished'),
+                    new ToolbarAction('sulu_admin.copy'),
+                ];
+
+                if (\count($locales) > 1) {
+                    $editDropdownToolbarActions[] = new ToolbarAction('sulu_admin.copy_locale');
+                }
+
                 $formToolbarActionsWithType[] = new DropdownToolbarAction(
```
rogamoore, line 242: I wasn't able to figure out the meaning of `formToolbarActionsWithType` and `formToolbarActionsWithoutType`, but so far it looks like this is the correct place.

rogamoore, line 241: I wasn't able to figure out the meaning of `formToolbarActionsWithType` and `formToolbarActionsWithoutType`, but so far it looks like this is the correct place.

Controller/ArticleController.php
```diff
@@ -528,7 +528,7 @@ public function postTriggerAction(string $id, Request $request): Response
     {
         // extract parameter
         $action = $this->getRequestParameter($request, 'action', true);
-        $locale = $this->getRequestParameter($request, 'locale', true);
+        $locale = $this->getRequestParameter($request, 'locale', false);
```
rogamoore, line 531: I had to change this to `false`, because the request does not provide any locale and it's working fine without locale. However, not sure if any other action requires the locale - in this case we would need to split the logic.

Tests/Functional/Controller/ArticleControllerTest.php
```diff
@@ -1193,6 +1193,29 @@ public function testCDelete()
         $this->assertNull($this->findViewDocument($article2['id'], 'de'));
     }
 
+    public function testCopy()
+    {
+        // create article
```
rogamoore, line 1198: I wasn't able to find any info on how to run the tests locally, so I blindly wrote this small test based on the existing ones.

alexander-schranz: You can run the tests via:

```
composer bootstrap-test-environment
composer test
```

A running elasticsearch instance and database is required.

PS: please your branch to `2.5` which is the next feature branch. Will have a deeper look at the pull request in the new year. Thx for working on this one 👍

rogamoore: I tried running the tests, but it's failing with the same error as on Github Actions:
```
Trying to register two bundles with the same name "ONGRElasticsearchBundle"
```
This doesn't seem to be related to my changes.

alexander-schranz: Can you try to repush. That issue should be gone after rebase to `2.5` that currently an issue on `2.x` which should not effect this branch.

rogamoore: Hm, I think I did that already. Any hints on how to repush? Or maybe the Pipeline can be retriggered?

rogamoore:
Just a quick note: I will be away for three weeks now so I won't be able to respond to feedback.

rogamoore:
I'm back, so if there is anything I can do to bring this forward please let me know.

alexander-schranz:
@rogamoore 

I repushed your branch. To retrigger a changed target branch pull request you need a new commit or change the commit you already pushed I mostly doing that via:

```bash
git commit --amend --date="now" # replace existing commit with anew data
git push myremote mybranch --force # hard push as we replaced the commit
```

It seems like some tests are failing now on lower PHP versions. Can you have a look at them why they are failing?

rogamoore:
@alexander-schranz Thank you!

I had a look - the tests are failing because as mentioned in the PR description right now it only works with Sulu >=2.6. We need this change https://github.com/sulu/sulu/pull/6859/files#diff-579b061b25841f1a028b4adc7b447e987e298306a441dd13346d08de3ec2d5ba to be backported to earlier Sulu versions  (which I think shouldn't be a BC break)

alexander-schranz:
@rogamoore would make sense to me to backport that line to 2.4 and 2.5. Do you have time to target that today, then I could add it to the next release this week?

rogamoore:
@alexander-schranz Sure, I will take care of it this afternoon

rogamoore:
See https://github.com/sulu/sulu/pull/7007

alexander-schranz:
The release is tagged so we can update the composer.json with `"sulu/sulu": "~2.4.11 || ^2.5.7"`.

rogamoore:
Tests are now green 🎉

rogamoore:
Just to confirm, there is nothing missing from my side, right?

alexander-schranz:
Thx for pinging, I just wanted to check it out and test it manually if all works like expected. Sorry for the late response here, I tried it out now and did run into the following issue:

Bildschirmfoto 2023-03-06 um 11 20 42

It could be related to that I have 2 languages, else all is like the default sulu installation with article bundle based on this branch 🤔 :

```xml
    <localizations>
        <!-- See: http://docs.sulu.io/en/latest/book/localization.html how to add new localizations -->
        <localization language="en" default="true"/>
        <localization language="de"/>
    </localizations>
```

Do we maybe need to use a fallback locale if there is no locale provided by the copy toolbar action or is it maybe required that the copy toolbar action also provides the current locale, the pages copy action seems currently sending a `locale=..` parameter.

I pushed my Repo here: https://github.com/alexander-schranz/reproducer-article-copy-error

alexander-schranz:
It seems to be related if there is route mapping defined for the ArticleDocument and live generation is used 🤔

rogamoore:
Thanks for testing! Sounds a bit like the problem described here: https://github.com/sulu/SuluArticleBundle/issues/538

alexander-schranz:
You are right with:

```yaml
sulu_route:
    mappings:
        Sulu\Bundle\ArticleBundle\Document\ArticleDocument:
            generator: schema
            options:
                route_schema: '/articles/{is_array(object) ? implode("-", object) : object.getTitle()}'
```

it works. We should document that in the UPGRADE.md. But maybe additional change we require to avoid running into that problem, will require to check that with @sulu/core-team.

rogamoore:
If the required changes are not too complicated, I could try to work on a PR.

alexander-schranz:
@rogamoore I debugged a little bit and prepared some changes in the sulu core: https://github.com/sulu/sulu/pull/7041 do you maybe want to test if this now works in the different cases?

rogamoore:
@alexander-schranz Sorry, I remember seeing the notification but somehow then forgot about it.
Looks promising, I will test it this afternoon!

alexander-schranz:
@rogamoore thank you!

rogamoore:
Thanks a lot! Can't wait for the release 👍

PR #611, opened 2022-05-26
Performance optimizations

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Performance optimizations
- When updating an article page, only persist parent article if necessary

### DIFF
--- a/Document/Subscriber/ArticlePageSubscriber.php
+++ b/Document/Subscriber/ArticlePageSubscriber.php
@@ -160,6 +160,10 @@ class ArticlePageSubscriber implements EventSubscriberInterface
             return;
         }
 
+        if (WorkflowStage::TEST === $document->getParent()->getWorkflowStage()) {
+            return;
+        }
+
         $document->getParent()->setWorkflowStage(WorkflowStage::TEST);
         $this->documentManager->persist(
             $document->getParent(),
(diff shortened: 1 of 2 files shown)


Document/Subscriber/ArticlePageSubscriber.php
```diff
@@ -160,6 +160,10 @@ public function setWorkflowStageOnArticle($event)
             return;
         }
 
+        if (WorkflowStage::TEST === $document->getParent()->getWorkflowStage()) {
```
wachterjohannes, line 163: @luca-rath can you try to explain me this?

luca-rath: @wachterjohannes The only thing this event handler does is setting the workflow stage of the parent document to `TEST`. But this is of course not necessary, if the workflow stage is already `TEST`. The problem with persisting is, that there's a listener listening on persist and updating all the child article pages, and if this can be omitted, it makes the process of saving (and importing in my case) quite a bit faster.

wachterjohannes: @luca-rath that makes sense :) but why did the tests fail? i know it looks like that it has nothing todo with your changes but we should fix it

luca-rath: @wachterjohannes didn't have time yet to investigate that further

PR #622, opened 2022-09-15
Add headless integration

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Add integration with SuluHeadlessBundle

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,6 +50,7 @@
         "phpstan/phpstan-webmozart-assert": "^1.0",
         "phpunit/phpunit": "^8.2",
         "sulu/automation-bundle": "^2.0@dev",
+        "sulu/headless-bundle": "^0.8.0@dev",
         "symfony/browser-kit": "^4.3 || ^5.0 || ^6.0",
         "symfony/dotenv": "^4.3 || ^5.0 || ^6.0",
         "symfony/framework-bundle": "^4.3 || ^5.0 || ^6.0",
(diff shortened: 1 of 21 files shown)


composer.json
```diff
@@ -18,7 +18,7 @@
         }
     ],
     "require": {
-        "php": "^7.2 || ^8.0",
+        "php": "^7.3 || ^8.0",
```
luca-rath, line 21: @alexander-schranz I did increase the min-requirement here, because I needed to require the headless-bundle as dev-dependency, which also requires `php >= 7.3` to be able to test the headless integration and I do not want to support php 7.2, if we don't run the tests on that php version

luca-rath: And since this will be released as a minor version (2.5), it should be fine I guess

alexander-schranz: fine for me.

alexander-schranz:
Would more go with `.../Infrastructure/Sulu/Headless` for the namespace, as we avoid the `Bundle` to be used as a namespace.

luca-rath:
> Would more go with `.../Infrastructure/Sulu/Headless` for the namespace, as we avoid the `Bundle` to be used as a namespace.

Okay, will fix this in a new PR then

PR #636, opened 2023-03-24
Fix use of the correct localizations with multiple portals

mario-fehr:
Use the correct localizations of the portal if multiple portals are configured and ´country´ key to the localizations array to match it with the other implementations in sulu.

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This fixes an issue with localizations when you split them in multiple portals.

#### Why?

Before this change all localizations of a webspace were returned even if you had separated them in multiple portals.

image

After this change, only localizations of the active portal are returned if you have multiple portals.

image

Also, the `country` key was added to match the output of pages.

### DIFF
--- a/Document/Serializer/WebsiteArticleUrlsSubscriber.php
+++ b/Document/Serializer/WebsiteArticleUrlsSubscriber.php
@@ -138,11 +138,18 @@ class WebsiteArticleUrlsSubscriber implements EventSubscriberInterface
             return;
         }
 
+        if (null !== ($portal = $attributes->getAttribute('portal'))) {
+            $allLocalizations = $portal->getLocalizations();
+        } else {
+            $allLocalizations = $webspace->getAllLocalizations();
+        }
+
         $urls = [];
         $localizations = [];
         $publishedLocales = $this->getPublishedLocales($article, $webspace);
 
-        foreach ($this->getWebspaceLocales($webspace) as $locale) {
+        foreach ($allLocalizations as $localization) {
+            $locale = $localization->getLocale();
             $published = \in_array($locale, $publishedLocales, true);
             $path = '/';
             $alternate = false;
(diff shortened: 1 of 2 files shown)


Document/Serializer/WebsiteArticleUrlsSubscriber.php
```diff
@@ -138,11 +138,18 @@ public function addUrlsOnPostSerialize(ObjectEvent $event): void
             return;
         }
 
+        if (null !== ($portal = $attributes->getAttribute('portal'))) {
+            $allLocalizations = $portal->getLocalizations();
```
alexander-schranz, line 142: @chirimoya we need to discuss this behaviour.

Document/Serializer/WebsiteArticleUrlsSubscriber.php
```diff
@@ -160,6 +167,7 @@ public function addUrlsOnPostSerialize(ObjectEvent $event): void
             $localizations[$locale] = [
                 'locale' => $locale,
                 'url' => $this->webspaceManager->findUrlByResourceLocator($path, null, $locale),
+                'country' => $localization->getCountry(),
```
alexander-schranz, line 170: this case seems to miss here also in sulu core: https://github.com/sulu/sulu/blob/b378ec72f296afedbf913967e5e66604737f4b48/src/Sulu/Bundle/WebsiteBundle/Resolver/TemplateAttributeResolver.php#L100

Document/Serializer/WebsiteArticleUrlsSubscriber.php
alexander-schranz: To be consistent to the pages and sulu core I'm fine to merge this currently this way.
Still it currently would not be possible to switch to other portals. My personally opinion would output always all localizations and give a hint also in which portals that locales appear.

PR #642, opened 2023-06-02
Respect webspaces when calculating the maximum number of sitemap pages

xabbuh:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs |
| License | MIT

#### What's in this PR?

When articles are collected for the sitemap, the provider restricts them to the articles that are associated with the webspaces of the current host. However, this restriction is not applied when the number of pages is being calculated.

#### To Do


TheCadien:
I think @xabbuh is right here and there is a bug. This should be taken into account.

TheCadien (approved):

alexander-schranz (approved):

alexander-schranz:
Thx @xabbuh

PR #656, opened 2023-12-21
Fix unexpected preview crash with adding preview condition 

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixes the preview when the locale is switched to a non existing one in the article detail view.

### DIFF
--- a/Admin/ArticleAdmin.php
+++ b/Admin/ArticleAdmin.php
@@ -260,6 +260,7 @@ class ArticleAdmin extends Admin
                 ];
             }
 
+            $previewCondition = 'locale in availableLocales';
             $viewCollection->add(
                 $this->viewBuilderFactory->createListViewBuilder(static::LIST_VIEW . '_' . $typeKey, '/:locale/' . $typeKey)
                     ->setResourceKey(ArticleDocument::RESOURCE_KEY)
@@ -308,6 +309,7 @@ class ArticleAdmin extends Admin
                     ->setTabCondition('shadowOn == false')
                     ->setTabPriority(1024)
                     ->addToolbarActions($formToolbarActionsWithType)
+                    ->setPreviewCondition($previewCondition)
                     ->setParent(static::EDIT_FORM_VIEW . '_' . $typeKey)
             );
             $viewCollection->add(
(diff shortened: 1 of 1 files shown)


Admin/ArticleAdmin.php
```diff
@@ -260,6 +260,8 @@ function(Localization $localization) {
                 ];
             }
 
+            $previewCondition = 'route != null';
```
Prokyonn: @alexander-schranz @wachterjohannes that was the best solution I was able to find for the preview condition 😅  it is working but somehow feels a bit wrong, do you have any better suggestion?

Admin/ArticleAdmin.php
```diff
@@ -260,6 +260,7 @@ function(Localization $localization) {
                 ];
             }
 
+            $previewCondition = "locale in availableLocales";
```
Prokyonn: @wachterjohannes Checking the locale is only possible, when we also update the [PreviewForm.js](https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/AdminBundle/Resources/js/views/PreviewForm/PreviewForm.js#L16) in sulu/sulu like this:

```diff
-    const enablePreview = !previewCondition || jexl.evalSync(previewCondition, this.resourceFormStore.data);
+    const previewData = {
+        ...this.props.router.attributes,
+        ...toJS(this.resourceFormStore.data)
+    };
+    const enablePreview = !previewCondition || jexl.evalSync(previewCondition, previewData);
```

alexander-schranz: is there not a `locale` and `originalLocale` returned by the JSON response?
When we add router attributes to the conditoin we maybe should prefix it with `__` like we did it with `__parent` or `__webspaces` already. So maybe `__routeAttributes`.

Prokyonn: <details>
  <summary>Response json</summary>

  image

</details>


Unfortunately we do not have a `locale` or `originalLocale` only the `ghostLocale` and this is not helpful.

So should we go with `__routeAttributes` ?

PR #679, opened 2024-06-13
Throw `NotAcceptableHttpException` if template request format does not exists

FlorianKoerner:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

A request format for which the template is not available throws a Twig\Error\LoaderError. An HttpException would be better.

See also:
- https://github.com/sulu/sulu/blob/2.6.2/src/Sulu/Bundle/WebsiteBundle/Controller/WebsiteController.php#L57-L59
- https://sulu.rocks/en/blog/a-great-song-will-win.invalid-request-format


alexander-schranz:
This make sense. Can you rebase to check if that fixes the PHP Lint errors?

alexander-schranz:
@FlorianKoerner Thank you!

PR #683, opened 2024-07-11
Add property for the routePathName

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/sulu/SuluPHPCRMigrationBundle/pull/2
| License | MIT

#### What's in this PR?

Adds a separate property for the routePathName.

#### Why?

The property is required for the migration to the SuluContentBundle with the SuluArticleBundle 3.0

### DIFF
--- a/Document/Subscriber/RoutableSubscriber.php
+++ b/Document/Subscriber/RoutableSubscriber.php
@@ -310,7 +310,7 @@ class RoutableSubscriber implements EventSubscriberInterface
 
         $node = $this->documentInspector->getNode($document);
         $node->setProperty($propertyName, $route->getPath());
-        $node->setProperty($this->propertyEncoder->localizedContentName(self::ROUTE_FIELD_NAME, $locale), $propertyName);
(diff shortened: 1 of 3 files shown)


Document/Subscriber/RoutableSubscriber.php
```diff
@@ -398,7 +401,7 @@ private function getRoutePathPropertyName(string $structureType, string $locale)
     {
         $metadata = $this->metadataFactory->getStructureMetadata('article', $structureType);
 
-        if ($metadata->hasTag(self::TAG_NAME)) {
```
Prokyonn: @alexander-schranz @wachterjohannes looks like this never worked. The `hasTag` method only validates the tags directly on the template, but not the tags from the properties 😅

Prokyonn: As discussed, we won't fix it

Prokyonn:
@alexander-schranz everything green now with symfony 7 compatibility 🥳

alexander-schranz (approved):

PR #698, opened 2024-08-22
Fix PhpCR migration exception for unpublished languages

Cephra:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR checks if the nodes have properties during the migration

#### Why?

Without this, articles which haven't been published cause migrations to fail.

### DIFF
--- a/Resources/phpcr-migrations/Version202407111600.php
+++ b/Resources/phpcr-migrations/Version202407111600.php
@@ -89,6 +89,9 @@ class Version202407111600 implements VersionInterface, ContainerAwareInterface
             /** @var Row<mixed> $row */
             foreach ($rows as $row) {
                 $node = $row->getNode();
+                if (!$node->hasProperty($templateKey)) {
+                    continue;
+                }
                 $structureType = $node->getPropertyValue($templateKey);
                 $routePathPropertyName = $this->getRoutePathPropertyName($structureType, $locale);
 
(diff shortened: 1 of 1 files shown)


Prokyonn (approved):

Prokyonn:
@Cephra Thank you for the merge request. The check for property existence does make sense, but I’m not entirely sure it applies to unpublished articles.

If an article is only saved as a draft and not published it should not be selected because it does not have a `mixinType` so the condition `([jcr:mixinTypes] = "sulu:article" OR [jcr:mixinTypes] = "sulu:articlepage")` should be false. After it was published once, the `template` property must be set. 

So I cannot really reproduce your issue 🤔

alexander-schranz:
@Cephra can you show us how such article looks like in your phpcr storage? How can such articles be created?

Prokyonn:
I was able to reproduce the issue. It occurs when you have multiple locales (e.g., `de` and `en`) and only publish the article in `de`. In this case, the unlocalized `mixin:type` property is set correctly, but for `en`, the `templateKey` is missing, causing the migration to crash.

Thanks to @Cephra for identifying and fixing this issue in a merge request!

Cephra:
Hello @Prokyonn & @alexander-schranz,

firstly thanks for accepting and merging the fix.

I also want to thank @FlorianKoerner who has helped me find and fix this bug!

PR #699, opened 2024-08-23
Add max_per_page filter to ArticleDataProvider

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

Adds the filter `max_per_page` to the smart content provider.


fnagel:
@alexander-schranz Do I understand this PR correctly when saying there is no way in v2.5 to use smart content with a pagination?

alexander-schranz:
@fnagel no `max_per_page` is possible in Sulu 1.x and Sulu 2.x already. This Pull Request implements max_per_page into the new 3.0 content repository. As documented the related documentations foreach version the param max_per_page existed before also in older content storage.

fnagel:
@alexander-schranz Ahh ok, thanks for the response. I will try some more and open a discussion if needed!

PR #712, opened 2024-12-02
Add clearing of different document manager to reindex command

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add clearing of different document manager to reindex command.

#### Why?

This should avoid usage of memory. Tested project did decrease memory usage from 970mib -> 480mib. I tried to also clearing the entityManager and disable the sqllogger interesting that ended in more usage of the Memory instead of less so I kept the sqllogger disabling out.

### DIFF
--- a/Command/ReindexCommand.php
+++ b/Command/ReindexCommand.php
@@ -199,10 +199,22 @@ class ReindexCommand extends Command
         $progressBar = new ProgressBar($output, $count);
         $progressBar->setFormat(' %current%/%max% [%bar%] %percent:3s%% %elapsed:6s%/%estimated:-6s% %memory:6s%');
         $progressBar->start();
+        $count = 0;
 
         foreach ($documents as $document) {
             $indexer->index($document);
             $progressBar->advance();
+
+            ++$count;
+
+            if (0 === ($count % 100)) {
+                $indexer->flush();
+                $this->documentManager->clear();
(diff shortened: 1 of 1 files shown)


Command/ReindexCommand.php
```diff
@@ -199,10 +199,22 @@ protected function indexDocuments(string $locale, IndexerInterface $indexer, Out
...
         $progressBar->start();
+        $count = 0;
 
         foreach ($documents as $document) {
             $indexer->index($document);
             $progressBar->advance();
+
+            ++$count;
+
+            if (0 === ($count % 100)) {
```
Prokyonn, line 210:
```suggestion
            if ($count !== 0 && 0 === ($count % 100)) {
```

to prevent flushing after the first

Prokyonn (changes requested):

Command/ReindexCommand.php
alexander-schranz: @Prokyonn no flush on first as ++$count is done before ;)

https://3v4l.org/rXKr8

Prokyonn: ahh yeah nvm 🙈

PR #719, opened 2025-09-11
Add content-fields to admin list search

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT

#### What's in this PR?

The `contentFields` should be included in the Admin-List search.


rhp-zellerd (approved):

Prokyonn (approved):

.github/workflows/test-application.yaml
```diff
@@ -21,19 +21,6 @@ jobs:
             fail-fast: false
             matrix:
                 include:
-                    - php-version: '7.3'
```
Prokyonn: as discussed, readd 7.3

rhp-geisela (approved):

alexander-schranz (approved):
Code changes looks fine.

If I understand it correclty this is a field which is searched but never shown? From my personally opinion would not recommend it to be enabled by default, making something searchable which isn't represented in a **List** view is a bad UX and ends in unexpected list results, thats why searches requires a special UI.

wachterjohannes:
> Code changes looks fine.
> 
> If I understand it correclty this is a field which is searched but never shown? From my personally opinion would not recommend it to be enabled by default, making something searchable which isn't represented in a **List** view is a bad UX and ends in unexpected list results, thats why searches requires a special UI.

@alexander-schranz you are totally right - we should fix that correctly for the 3.0 version. But for now thats the best solution

PR #724, opened 2026-03-18
Fix routePathName subscriber

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | yes
  | New feature? | no
  | BC breaks? | no
  | Deprecations? | no
  | License | MIT

  #### What's in this PR?

  - Added new `RoutePathNameSubscriber` that writes the `routePathName` property on article PHPCR nodes during persist and publish
  - Added PHPCR migration `Version202603181200` to backfill `routePathName` on all existing article nodes (both default and live sessions)
  - Added functional tests verifying `routePathName` is written on persist and publish

  #### Why?

  The `routePathName` property was introduced in #683 to store the encoded route path property name on article PHPCR nodes, which is needed by the PHPCR migration bundle to resolve the correct route URLs during
  migration.

  However, the property was not written because the `RoutableSubscriber` (which contained the write logic) had its service registration removed somehow.

  This PR fixes the issue by introducing a dedicated `RoutePathNameSubscriber` that solely handles writing the `routePathName` property. A new PHPCR migration re-runs the backfill for all existing articles since
  the original migration had no effect.

  #### To Do

  - [ ] Create a documentation PR

### DIFF
--- a/Document/Subscriber/RoutePathNameSubscriber.php
+++ b/Document/Subscriber/RoutePathNameSubscriber.php
@@ -1,0 +1,82 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ArticleBundle\Document\Subscriber;
+
+use Sulu\Bundle\ArticleBundle\Document\Behavior\RoutableBehavior;
+use Sulu\Bundle\DocumentManagerBundle\Bridge\PropertyEncoder;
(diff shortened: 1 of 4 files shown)


Controller/ArticleController.php
```diff
@@ -586,7 +586,7 @@ public function deleteAction(Request $request, string $id): Response
         $deleteLocale = $this->getBooleanRequestParameter($request, 'deleteLocale', false, false);
 
         $document = $this->documentManager->find($id);
-        if ($deleteLocale) {
+        if ($deleteLocale && null !== $locale) {
```
wachterjohannes: is this intended?

wachterjohannes (changes requested):

wachterjohannes (approved):

